PR #7, opened 2010-12-02
Bz107 etag quoting

dizzyd:
Please review this fix for bz107.


jonmeredith:
Scott beat me to it - filed a bug here with similar comments https://issues.basho.com/show_bug.cgi?id=912

PR #8, opened 2010-12-22
Bz884 etag list

dizzyd:
Implemented support for 884, etag lists. When reviewing, please also verify that original reported curl usage is resolved. I also added a QC test that should provide more thorough validation.


justinsheehy:
+1

PR #9, opened 2011-01-28
Fixing bash path in new_webmachine.sh

l1x:
The proper way to invoke bash is the following:
/usr/bin/env bash

It works on linux, *bsd, macos and more :)


l1x:
Using the portable way of invoking bash.

PR #10, opened 2011-01-28
include eunit headers only when necessary

lemenkov:
include eunit only within the ifdef(TEST) blocks. This avoids an eunit dependency during non-test compilation.


dizzyd:
Merged to mainline.

PR #13, opened 2011-03-16
Drop useless include

lemenkov:
it's safe to remove the inclusion of this header - it only contains definition of the ?RECBUF_SIZE macro which isn't used anywhere in webmachine.


dhull:
+1. This include also breaks edoc creation with the error:

```
./src/webmachine_request.erl: at line 81: can't find include lib "mochiweb/include/internal.hrl"
edoc: skipping source file './src/webmachine_request.erl': {'EXIT',error}.
```

PR #17, opened 2011-03-25
Fix for default app template crashing (at least on Erlang R13B03)

capotej:
Couldn't get a demo app running from a fresh new_machine.sh, googling the error found a fix here: https://gist.github.com/878868

(awesome project btw!)

### DIFF
--- a/priv/templates/src/wmskel.app.src
+++ b/priv/templates/src/wmskel.app.src
@@ -8,6 +8,7 @@
   {applications, [
                   kernel,
                   stdlib,
+		  inets,
                   crypto,
                   mochiweb,
                   webmachine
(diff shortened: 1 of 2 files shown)


capotej:
No problem, updated as requested.

dylanz:
Would love to get this in there, ran into the same issue.  Made me look at other frameworks for a little while ;)

priv/templates/src/wmskel.app.src
```diff
@@ -8,6 +8,7 @@
   {applications, [
                   kernel,
                   stdlib,
+		  inets,
```
Vagabond: Fix the indenting on this line to use spaces, not tabs.

Vagabond:
If you fix the whitespace error noted above, I think this looks good to merge.

capotej:
Fixed, my bad.

PR #18, opened 2011-04-27
Adding support for more mime types.

etnt:
Hi, 

I've added a dependency to a package named moo_mime_type and is making use of it
in the webmachine_util:guess_mime/1 code. The moo_mime_type code is basically
stolen from yaws and generates the moo_mime_types.erl file based on a mime.types
file, originally taken from Apache. Any updates with new mime types is done in this file
and a new moo_mime_types.erl file is generated and checked in. In case the moo_mime_types
code should fail (?), the old webmachine guss_mime code is used as a fallback.

Btw: the old code returned 'text/plain' as default, perhaps is should return: 'application/octet-stream' ?

Cheers, Tobbe

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -3,6 +3,12 @@
 {cover_enabled, true}.
 
 {deps, [
-        {mochiweb, "1.5.1", {git, "git://github.com/mochi/mochiweb",
-                            {tag, "1.5.1"}}}
+        {mochiweb, "1.5.1",
+         {git, "git://github.com/mochi/mochiweb",
+          {tag, "1.5.1"}}}
+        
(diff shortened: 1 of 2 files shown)


etnt:
Is pull requests dealt with? I mean is it any meaning in making them? An 'accept', 'reject' or 'pending' would have been nice...

beerriot:
Sorry for the delay, Tobbe.  At this time, we'd rather not add this dependency to Webmachine.  If an app developer prefers to use moo_mime_types, they should add the dependency to their app, and write their resources to call it instead of `webmachine_util:guess_mime`.

Thank you for raising awareness of a useful library, though.

-Bryan

PR #19, opened 2011-05-09
Fix edoc errors (macros and includes)

beerriot:
As Björn reported (http://lists.therestfulway.com/pipermail/webmachine_lists.therestfulway.com/2011-May/000505.html), edoc didn't like wmtrace_resource's use of macros.  After clearing that up, edoc didn't like some of the include defs.  These two commits should fix both problems.


justinsheehy:
+1

PR #20, opened 2011-05-18
removes proplists in favor of lists:keyfind for performance

roder:
This morning while watching bashochat, I saw that andrew said "switching proplists to lists is a significant win", which encouraged me to submit this pull request.

About 6 months ago  I switched all the proplists:get_value to lists:keyfind.  I never issued a pull request because I thought it was trivial, but given the chat this morning I was inspired to finally submit it.  I made sure it was all up-to-date with basho/webmachine/tree/master so it should merge nicely.

Thanks.

### DIFF
--- a/src/webmachine_mochiweb.erl
+++ b/src/webmachine_mochiweb.erl
@@ -111,7 +111,10 @@ loop(MochiReq) ->
     end.
 
 get_option(Option, Options) ->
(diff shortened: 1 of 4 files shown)


beerriot:
Any numbers on how much these patches affect speed?

PR #26, opened 2011-06-08
Automatically insert a Location header as appropriate

b:
POST create should automatically insert a Location header, but currently does not.  This branch adds that functionality without breaking existing code.  There are two behavior changes:
- If the string returned from create_path/2 started with "/" it is treated as an absolute path and is not simply appended to the dispatch path.
- If base_uri/2, a new resource function, returns a string, it is used as the prefix for the path information in the Location header.  If base_uri/2 returns undefined (the default), then a base URI is automatically constructed from the information in the request.


justinsheehy:
Reviewed, and also discussed with Bryan.

This behavior is desired, appears to be correctly implemented, and should have minimal impact on existing resources.

(old wmtrace debug logs will not work due to a record change, but that's minimal)

Merging.

PR #28, opened 2011-06-29
AZ460 - Update webmachine dispatch list to support guard function.

rustyio:
Update webmachine to dispatch guards. Can now use path specs of the form `{PathSpec, Guard, Mod, Options}` where Guard is either a function or a {Mod, Fun} tuple. The function has arity 1. Webmachine runs the guard, passing in the request object as the sole parameter. The route can only match if the function returns true.

### DIFF
--- a/src/webmachine_dispatcher.erl
+++ b/src/webmachine_dispatcher.erl
@@ -21,7 +21,7 @@
 -author('Justin Sheehy <justin@basho.com>').
 -author('Bryan Fink <bryan@basho.com>').
 
--export([dispatch/2, dispatch/3]).
+-export([dispatch/3, dispatch/4]).
 
 -define(SEPARATOR, $\/).
 -define(MATCH_ALL, '*').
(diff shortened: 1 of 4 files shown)


src/webmachine_dispatcher.erl
```diff
@@ -170,18 +181,48 @@ bind_port(PortAtom, Port, Bindings) when is_atom(PortAtom) ->
...
     end.
 
+run_guard(undefined, _RD) ->
+    true;
+run_guard(Fun, RD) when is_function(Fun) ->
+    try
+        Fun(RD) == true
+    catch _Type : Msg ->
+            error_logger:error_msg("Error running guard ~p: ~p~n", [Fun, Msg]),
+            throw({error_running_guard, Fun, Msg})
```
beerriot, line 213: Failing in this manner results in an empty reply from the server.  Either of these alternates seem valid to me:
1. Immediate 500 response.
2. Simply fail the match, and move on to the next dispatch entry.

rustyio:
Opted for the 500 response. 

Updates:
1. Wrapped the webmachine_dispatch call in a try/catch block. 
2. Moved normal case code around (webmachine_dispatch lines 80-86) in order to properly handle cases where module is missing, or where module doesn't have an init function. Previously returned an empty reply from the server.
3. Consolidated existing error handling code in webmachine_dispatch.

beerriot:
+1 looks good to me, thanks!

PR #30, opened 2011-07-05
Validate Content-MD5 header

seancribbs:
Adds validation of the Content-MD5 header when submitted with a request body.
- Returns `400 Bad Request` when Content-MD5 header doesn't match checksum of request body.
- Adds optional `validate_content_checksum` resource callback. Resource can return `false` if the checksum fails, `not_validated` (default) to use the default validation, or anything else (e.g. `true` or `ignore`) to continue normal flow.
- Adds generic error rendering function to `webmachine_error_handler`.

Example of implicit usage: https://gist.github.com/584d6b86ec58f83ecba5

TODO: 
- Handles both plain and streamed/chunked responses, but sets the request body state after streaming (might break some things?). Please review.

### DIFF
--- a/src/webmachine_decision_core.erl
+++ b/src/webmachine_decision_core.erl
@@ -178,8 +178,35 @@ decision(v3b10) ->
                    string:join([atom_to_list(M) || M <- Methods], ", ")}]}),
             respond(405)
     end;
-%% "Malformed?"
+
+%% "Content-MD5 present?"
 decision(v3b9) ->
+    decision_test(get_header_val("content-md5"), undefined, v3b9b, v3b9a);
+%% "Content-MD5 valid?"
+decision(v3b9a) ->
+    case resource_call(validate_content_checksum) of
+        {error, Reason} ->
+            error_response(Reason);
+        {halt, Code} ->
+            respond(Code);
+        not_validated ->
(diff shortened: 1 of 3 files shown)


refactor:
It looks you misunderstand HTTP Content-MD5 header, Context-MD5 value should be BASE64 encoded chars on md5 checksum, rather than the hex numbers of md5 checksum. You can check:
"The Content-MD5 Header Field"
http://www.ietf.org/rfc/rfc1864.txt
and
http://en.wikipedia.org/wiki/List_of_HTTP_headers

reiddraper:
This does look like a misunderstanding. @seancribbs @vinoski I'm happy to make a PR to correct, just want to make sure I'm not missing something first?

vinoski:
I don't think you're missing anything; I believe a correction is needed.

reiddraper:
There's also a bug where the code says `md5update`, where it should be `md5_update`.

reiddraper:
Tracking down (what is likely my own bug) where I'm getting `{error,{exit,"wrq:req_body and wrq:stream_req_body conflict"`, but will tackle this fix once I get past that.

seancribbs:
I agree with @vinoski.

PR #33, opened 2011-08-03
AZ564 - Update webmachine guard to include matched path information

rustyio:
We recently added "guard functions" to the webmachine dispatch list, which means you could specify a function to run as a final step after all of the hostname and path matching. (If the function returns true, then the route is an acceptable match.)

The guard function takes a single parameter, the `#wm_reqdata` record. Previously, we did not update the `#wm_reqdata` object with hostname or path tokens before calling the guard function.

This pull request adds code to update `#wm_reqdata` record before calling the guard. This allows the user to  call functions like `wrq:path_info(token, RD)`, just like they would in the webmachine resource, with the expected results.


beerriot:
Please add a test that checks that the hostname and path tokens are set in the reqdata passed to the guard.  +1 other than that.

PR #35, opened 2011-08-15
host_tokens in wm_reqdata is reversed

ulfa:
In the locationheader created by webmachine, the host_token is reversed. Example : 1.0.0.127 and not 127.0.0.1
To fix this problem i inserted a lists:reverse in line 129, but i think that it must be fixed in the root.

### DIFF
--- a/src/webmachine_request.erl
+++ b/src/webmachine_request.erl
@@ -124,7 +124,7 @@ x_peername(Default) ->
 call(base_uri) ->
     RD = ReqState#wm_reqstate.reqdata,
     Scheme = erlang:atom_to_list(RD#wm_reqdata.scheme),
-    Host = string:join(RD#wm_reqdata.host_tokens, "."),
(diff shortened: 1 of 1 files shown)


aerosol:
Hi, 

looks like unnecessary (?) `reverse` at: https://github.com/basho/webmachine/blob/master/src/webmachine_dispatcher.erl#L51

nyarly:
Actually, that reverse is (obscurely) necessary - host based dispatches of the form ['*', "example", "com"] should match not only www.example.com but also very.deeply.nested.example.com and collecting all the sub-sub domain parts would be much more difficult if the host part is un-reversed.

But, ulfa's change would be much appreciated.  Can we get this PR merged soon?

reiddraper:
Need to make sure to revert [this](https://github.com/basho/riak_kv/pull/371) (and it's 1.2 counter PR) when we merge this.

beerriot:
I believe the root fix to this issue is PR #87. I'm closing this PR to continue discussion there.

PR #37, opened 2011-08-18
Fix reversed hostname tokens on calling call(base_uri)

jvliwanag:
Host tokens are stored reversed. When forming the string for the base_uri, this should be reversed back.


PharkMillups:
Hey Jan, 

Thanks for sending this over. It looks like @ulfa happened to find this just a few days back and has already sent a pull request for it. 

https://github.com/basho/webmachine/pull/35 

I'm going to close this pull request as a duplicate. That said, thanks for sending it our way and keep the code coming!

Mark

PR #38, opened 2011-08-24
minor change to wmskel to make it compatible with erlang releases

loxs:
change how dispatch.conf is located, in order to make webmachine projects work with erlang releases

When you do an erlang release, ebin and priv directories do not end up in the same parent directory. As a result of that, the previous method of locating dispatch.conf fails, and webmachine crashes.


beerriot:
Looks like @vinoski reimplemented this change in https://github.com/basho/webmachine/commit/b544a32c59520340d64da0202c2cd9fd0295d977. Sorry for missing it, @loxs .

PR #39, opened 2011-09-07
Better visibility into errors during create_path handling in POST

d2fn:
expose underlying problem when create_path returns something other than a path string


justinsheehy:
+1

Looks fine to me, will merge this soon.

PR #40, opened 2011-09-13
Two simple edoc fixes.

beerriot:
Just cleaning up mismatching @specs and invalid @types.


dhull:
+1.  I just independently submitted a fix for these same edoc problems which I will withdraw in favor of your (superior) fix.

kellymclaughlin:
+1 to merge.

PR #42, opened 2011-09-28
Added popular MIME types

kardan:
Added support for popular mime types such as webfonts, mp4, webm and ogg. Sorted the types based on file name extension.

I ran into problems when using webfonts and thought it would be useful for everyone. Saw that there was more pull requests related to mime types that went deeper but thought it might be interesting anyway. I have used https://github.com/paulirish/html5-boilerplate/blob/master/.htaccess as guide. I did not want to get into a religious talk over the JavaScript(application/x-javascript) entry but it's worth mentioning that they are different.


russelldb:
It's all good, many thanks.

PR #43, opened 2011-09-30
Crash when streaming a body back to the client.

Mistagrooves:
I was attempting to stream a request back to the client with a {stream, streambody()} response. This resulted in a crash in webmachine_logger when calling integer_to_list with the atom 'chunked'.

=ERROR REPORT==== 30-Sep-2011::11:24:35 ===
*\* Generic server webmachine_logger terminating 
*\* Last message in was {'$gen_cast',
                        {log_access,
                         {wm_log_data,file_srv,
                          {1317,396275,125864},
                          'HEAD',
                          {4,
                           {"user-agent",
                            {'User-Agent',
                             "curl/7.21.4 (i386-pc-win32) libcurl/7.21.4 OpenSSL/0.9.8r zlib/1.2.5 libidn/1.18 libssh2/1.2.7 librtmp/2.3"},
                            {"host",
                             {'Host',"localhost:8000"},
                             {"accept",{'Accept',"_/_"},nil,nil},
                             nil},
                          "127.0.0.1","/files/sqlncli.msi",
                          {1,1},
                          200,chunked,
                          {1317,396275,172860},
                          {1317,396275,172861},
                          []}}}
*\* When Server state == {state,
                            {2011,9,30,15},
                            "priv/log/access.log",
                            {webmachine_logger,
                                "priv/log/access.log.2011_09_30_15",
                                {file_descriptor,prim_file,
                                    {#Port<0.3555>,720}}}}
*\* Reason for termination == 
*\* {badarg,[{erlang,integer_to_list,[chunked]},
            {webmachine_logger,format_req,1},
            {webmachine_logger,handle_cast,2},
            {gen_server,handle_msg,5},
            {proc_lib,init_p_do_apply,3}]}

### DIFF
--- a/src/webmachine_logger.erl
+++ b/src/webmachine_logger.erl
@@ -98,6 +98,11 @@ maybe_rotate(State, Time) ->
             State#state{hourstamp=ThisHour, handle=Handle}
     end.    
 
+format_length(chunked) ->
+	"chunked";
+format_length(C) ->
+	integer_to_list(C).
+	
 format_req(#wm_log_data{method=Method, 
                         headers=Headers, 
                         peer=Peer, 
@@ -108,7 +113,7 @@ format_req(#wm_log_data{method=Method,
     User = "-",
     Time = fmtnow(),
     Status = integer_to_list(ResponseCode),
-    Length = integer_to_list(ResponseLength),
+    Length = format_length(ResponseLength),
     Referer = 
         case mochiweb_headers:get_value("Referer", Headers) of
             undefined -> "";


reiddraper:
@Mistagrooves I've been unable to reproduce the crash. Would you mind posting some steps to reproduce?

Mistagrooves:
Sure, this stemmed from trying to return a HEAD request on a static file without having to read the entire file's contents in, just provide the size of the file in the content-length header. The problem was that providing an empty binary in the response to the content_types_provided Handler function resulted in the content-length header being overridden with 0.  

The work around that I found was to provide a streaming body that was essentially empty (which I admit is probably not to spec). This triggers the issue:

``` erlang
init([]) ->
    {ok, undefined}.

allowed_methods(ReqData, Context) ->
    {['HEAD'], ReqData, Context}.

to_html(ReqData, State) ->
    case wrq:method(ReqData) of
        'HEAD' ->
            ReqData2 = wrq:set_resp_header("Content-Length", "0", ReqData),
            {{stream, {<<>>, done}}, ReqData2, State}
    end.
```

dispatch.conf:

```
{["test"], test, []}.
```

Then a simple: `httpc:request(head, {"http://localhost:8000/test", [{"Accept", "text/html"}]}, [], []).` should trigger it.

reiddraper:
@Mistagrooves I understand your use-case with returning large files, but this is not currently possible with webmachine. There are several content-negotiation things like content-encoding that could change the content-length, and won't be known unless you actually produce (but not necessarily return) the body.

Mistagrooves:
@reiddraper Ok, I understand. The issue was less about my actual use case which this popped up in and more about the logging message that was produced.

reiddraper:
@Mistagrooves somewhat unrelated, I made a [webmachine branch](https://github.com/basho/webmachine/tree/wrd-content-length-with-streaming) that allows you to send streams with ahead-of-time known content length, and without doing `Transfer-Encoding: chunked`. It's not production ready, nor do I intend for it to be merged into master, but it's worth taking a look at.

Mistagrooves:
@reiddraper Neat, I'll take a look. Thanks.

reiddraper:
A year later..., [here's a pull request](https://github.com/basho/webmachine/pull/118) that addresses some of these issues.

kellymclaughlin:
The known content length work previously referenced by @reiddraper has been merged to `master` now so I'm going to go ahead and close this out.

PR #46, opened 2011-11-13
Add wrq:base_uri/1 and tests, also fixes bug in location URI generation

seth:
Hi,

This branch adds a convenience function wrq:base_uri to make it easier for implementors who want to include a URI in a generated response.

It also addresses an issue with the base_uri recently added to set the location header and teaches the dispatch code to use 443 as the default port for https.
- seth


ddosia:
Very usefull ability. If i could vote then +1 for this pull request.

beerriot:
I've rebased this as PR #88. I'm closing this issue to continue discussion there

PR #47, opened 2011-11-25
BZ1293: Fix filter_by_resource to handle the 4-tuple format that includes a predicate fun.

seancribbs:
When using the 4-tuple route format that includes a predicate function as the second element, no function clause of `filter_by_resource/2` will match, which makes it impossible to remove resources from the dispatch list using `remove_resource/1`.

This patch addresses the issue by adding additional clauses to `filter_by_resource/2` to handle the 4-tuple format.


beerriot:
This works for regular dispatch lines (thank you!), but it perpetuates a long-standing error in filtering host-based dispatch lines.  See https://github.com/basho/webmachine/pull/70 for the additional fix.

PR #48, opened 2011-11-28
Redirection support via rewrite_module configuration option

argv0:
Basic support for path redirection via a {webmachine, rewrite_module} configuration option:

rewrite_module should export a single function, rewrite/2 with the following signature:

rewrite(Headers::mochiweb_headers(), Path::string()) -> NewPath::string().

### DIFF
--- a/src/webmachine_mochiweb.erl
+++ b/src/webmachine_mochiweb.erl
@@ -56,7 +56,15 @@ start(Options) ->
     end,
     application_set_unless_env(webmachine, dispatch_list, DispatchList),
     application_set_unless_env(webmachine, error_handler, ErrorHandler),
-    application_set_unless_env(webmachine, rewrite_module, RewriteModule),
+    case RewriteModule of
+        undefined ->
+            %% webmachine:new_request/2 will explode if
+            %% application:get_env returns {ok, undefined}
+            ok;
+        _ ->
+            application_set_unless_env(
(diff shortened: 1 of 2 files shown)


src/webmachine_mochiweb.erl
```diff
@@ -49,13 +50,14 @@ start(Options) ->
...
-      {undefined, _} -> {?MODULE, Options4};
-      {PN, O5} -> {PN, O5}
+    {PName, Options6} = case get_option(name, Options5) of
+      {undefined, _} -> {?MODULE, Options5};
+      {PN, O6} -> {PN, O6}
     end,
     application_set_unless_env(webmachine, dispatch_list, DispatchList),
     application_set_unless_env(webmachine, error_handler, ErrorHandler),
-    mochiweb_http:start([{name, PName}, {loop, fun loop/1} | Options5]).
+    application_set_unless_env(webmachine, rewrite_module, RewriteModule),
```
beerriot: This call means that if no `rewrite_module` is defined in `Options`, then `rewrite_module` app env var is defined as `undefined`.  This is a problem for the case statement at the top of `webmachine:new_request`, as `application:get_env(webmachine, rewrite_module)` returns `{ok, undefined}`, not just `undefined`.

vinoski:
Merged manually.

PR #49, opened 2011-11-29
Fix support for media types with parameters

beorn247:
Using media types with parameters ("audio/vnd.wave; codec=31" for example) does not work. These commits will fix that, at least for GET requests.


kellymclaughlin:
I apologize for the ridiculously long time this has sat without action or comment. We have merged in [this](https://github.com/basho/webmachine/pull/112) pull request to address the same issue so I am going to close this one. Thank you for the issue report and pull request though!

PR #50, opened 2011-12-08
Suppress the error generated when webmachine_request does an exit(normal)

Vagabond:
This prevents a very common error (when you're doing a lot of chunked uploads) from filling up the logs.

A good way to reproduce the error is to do something like the following:

curl -v -d @somefile -H "Content-Type: text/plain" http://127.0.0.1:8098/riak/test --limit-rate 1  -H "Transfer-Encoding: chunked"

Where 'somefile' is some medium sized file. Then, while it's uploading, hit ctrl-c and you'll get the error.


reiddraper:
+1

PR #52, opened 2011-12-21
Added support for Range header when returning a non-stream / non-file body

srstrong:


src/webmachine_deps.erl
```diff
@@ -57,7 +57,19 @@ new_siblings(Module) ->
...
-    code:clash(),
+
+    case lists:filter(fun(Path) -> case lists:reverse(Path) of
+				       "tinue." ++ _ -> true;
+				       _ -> false
+				   end
+		      end, code:get_path()) of
+	[] ->
+	    code:clash();
+	_EUnitPath ->
```
seancribbs, line 68: Why not use -ifdef(TEST) instead?

srstrong, line 68: I could have done that, except it would have meant rebuilding webmachine with TEST defined whenever I want to test my system; either that, or keep two sets of beams and have to mess with the code path to pick up the correct one.  Neither appeals much, plus I don't want to treat webmachine any differently to all my other (numerous!) dependencies.

This looked like one way (albeit not perfect) of having the webmachine runtime detect if it was running in a test environment, and to hence skip the code:clash() call that otherwise was generating a number of warnings are the tests started running.

kellymclaughlin:
The changes in the request have all been addressed in other commits, so closing. Thanks.

PR #53, opened 2011-12-30
processable_entity decision

jamesgolick:
I wanted a place to check parameters and 422 from. I don't know the webmachine source at all, so I'm not sure whether I put this in the right place, or whether my decision IDs make any sense to you guys, so let me know.


vinoski:
While this code change is straightforward, at this stage we prefer not to make changes to the graph/flowchart. Also, 422 isn't a standard HTTP status code (see [RFC 7231](http://tools.ietf.org/html/rfc7231)).

PR #54, opened 2012-01-29
added necessary dep mochiweb to app file

tsloughter:


tsloughter:
Any chance this can be merged in soon? Its very small change but makes generating releases and target systems much easier because they then know to include mochiweb.

beerriot:
Sorry we ignored this request for so long.  It's now merged to master.  Thank you!

PR #55, opened 2012-02-17
Slf fix media_type_to_detail

slfritchie:
Strip whitespace from content-type (and others) parsed by `webmachine_util.erl:media_type_to_detail/1`.
Without this fix, Webmachine cannot handle this (AFAICT completely RFC 2616 compliant) header:

```
content-type:          application/x-www-form-urlencoded          ;      charset      =       utf8
```


kellymclaughlin:
+1 to merge.

PR #56, opened 2012-02-17
Strip whitespace from content-type (and others) parsed by webmachine_uti...

slfritchie:
Strip whitespace from content-type (and others) parsed by `webmachine_util.erl:media_type_to_detail/1`.
Without this fix, Webmachine cannot handle this (AFAICT completely RFC 2616 compliant) header:

```
content-type:          application/x-www-form-urlencoded          ;      charset      =       utf8
```


vinoski:
The body of the `webmachine_utils:media_type_to_detail/1` fun could be replaced with just a call to `mochiweb_util:parse_header/1` &mdash; even the return type is compatible. Plus it solves the problem this commit solves.

vinoski:
Closing this PR in favor of PR #94.

PR #58, opened 2012-02-25
Update `WMVSN` version to be aligned with git tag

djui:
The `WMVSN` version was not updated when a new tag was created. Fix it.


kellymclaughlin:
Sorry it took so long to address this. The lastest tag is now `1.9.2` so I'm closing this pr in favor of [this](https://github.com/basho/webmachine/pull/86) one. Thanks for the submission.

PR #63, opened 2012-03-02
Do not exit when mochiweb_socket:send returns an error (Known-size-stream branch)

beerriot:
This patch fixes issue #59.

Please see https://github.com/basho/webmachine/pull/62 for more information.  This patch is a rebased version of that one, with the same fix applied to the `known_size_stream` body producer.


kellymclaughlin:
I posted this in the chat, but realized I missed you so I'll post my message here too:

> bryan: just realized that wm change has the side effect of breaking how we detect dead wm procs in order stop fetching blocks on a get. that also messes with the accuracy of the billing info.

kellymclaughlin:
Closing because this branch has been merged to `master`.

PR #65, opened 2012-03-03
swap - to _ for app name.

JakubOboza:
micro change to rename app name like kuba-app to kuba_app so the generated code will be valid and compile.


vinoski:
Closing this in favor of PR #95, which handles more project name issues.

PR #66, opened 2012-03-22
Various Media Range Params Fixes

jrwest:
This follows up on a bug originally adressed by this pull request https://github.com/basho/webmachine/pull/21 but that I don't think was ever completed from what I can tell (I haven't been able to find any other references and I was able to recreate the issues mentioned).

2 bug fixes and a minor "improvement":
- Fixes formatting of content types that contain parameters. Originally, this would cause webmachine to crash in certain cases
- fixing above caused webmachine to crash in v3o18 instead. this no longer occurs
- It was documented in `webmachine_util:choose_media_type/2` that atoms could be used as keys in the media-types params list, this is now supported.

### DIFF
--- a/src/webmachine_decision_core.erl
+++ b/src/webmachine_decision_core.erl
@@ -528,7 +528,7 @@ decision(v3o18) ->
                               calendar:universal_time_to_local_time(Exp))})
             end,
             F = hd([Fun || {Type,Fun} <- resource_call(content_types_provided),
-                           CT =:= Type]),
+                           CT =:= webmachine_util:format_content_type(Type)]),
             resource_call(F);
         false -> nop
     end,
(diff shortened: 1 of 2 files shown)


kellymclaughlin:
How about a test case like this? (or even exactly this):

```
format_content_type_test() ->
    Types = ["audio/vnd.wave; codec=31",
             "Text/x-Okie; charset=iso-8859-1; declaration=\"<f950118.AEB0@XIson.com>\""],
    [?assertEqual(Type, format_content_type(
                          webmachine_util:media_type_to_detail(Type)))
     || Type <- Types],
    ?assertEqual(hd(Types), format_content_type("audio/vnd.wave", [{codec, "31"}])).
```

kellymclaughlin:
I wondered if it should support the value of a parameter pair being an integer, but I do not see that it can ever be converted to an integer so I don't think it's worth it.

jrwest:
added suggested test

jrwest:
closing this PR in favor of #112 (same branch rebased on master)

PR #68, opened 2012-04-30
App skeleton priv dir

vinoski:
This changes how the application priv directory is found in the application skeleton code. It now uses code:priv_dir instead of creating a priv dir pathname relative to the ebin directory.

The second commit just adds the .eunit dir to .gitignore.

### DIFF
--- a/priv/templates/src/wmskel_sup.erl
+++ b/priv/templates/src/wmskel_sup.erl
@@ -42,9 +42,9 @@ upgrade() ->
 %% @doc supervisor callback.
 init([]) ->
     Ip = case os:getenv("WEBMACHINE_IP") of false -> "0.0.0.0"; Any -> Any end,
-    {ok, Dispatch} = file:consult(filename:join(
-                         [filename:dirname(code:which(?MODULE)),
-                          "..", "priv", "dispatch.conf"])),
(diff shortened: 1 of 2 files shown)


priv/templates/src/wmskel_sup.erl
```diff
@@ -42,9 +42,10 @@ upgrade() ->
 %% @doc supervisor callback.
 init([]) ->
     Ip = case os:getenv("WEBMACHINE_IP") of false -> "0.0.0.0"; Any -> Any end,
+    {ok, App} = application:get_application(?MODULE),
     {ok, Dispatch} = file:consult(filename:join(
-                         [filename:dirname(code:which(?MODULE)),
-                          "..", "priv", "dispatch.conf"])),
+                                    [code:priv_dir(App),
+                                     "..", "priv", "dispatch.conf"])),
```
beerriot: Why `priv_dir()/../priv` instead of just `priv_dir()`?

vinoski: Umm, just making sure you're reviewing carefully? :)

Thanks for catching that, fixed.

beerriot:
:+1: merge; thanks, steve

PR #70, opened 2012-05-06
Properly filter per-host dispatches in webmachine_router:remove_resource/1

beerriot:
It looks like webmachine_router:remove_resource/1 never worked on
host-based dispatch lines because it was attempting to match {Host,
Path}, instead of {Host, [Path, ...]}.  This patch changes the filter to
delve into those per-host dispatch lists to apply the filter correctly
there as well.


kellymclaughlin:
Review by means of test cases. If you're good with my tests, I'm +1 to merge this to `bz1293-remove-resource-with-predicate` and for merging that to `master`.

beerriot:
Indeed, thanks @kellymclaughlin !

PR #73, opened 2012-05-17
Fix for trailing CRLF 

bjnortier:
Fix for trailing CRLF as per the mailing list:
http://lists.therestfulway.com/pipermail/webmachine_lists.therestfulway.com/2010-August/000272.html

I've edited the mailing list patch to use binary pattern pattern matching


kellymclaughlin:
Thanks for the fix. Sorry for the delay on looking at this, but I'm merging it in now.

bjnortier:
woop!

PR #74, opened 2012-05-24
Add Documentation of How to Write a Logger for Webmachine

rmeritz:
This adds a template logger as priv/template/logger.erl do provide some documentation of how to write your own logger.


kellymclaughlin:
I'm going to close this because the logging has changed on `master` such that logger modules are now event handlers instead of servers and this is no longer applicable.

PR #75, opened 2012-05-25
Exposed get_routes() to the public.

davidcoallier:
- Added the ability to use get_routes (Added to exports())
- Did not change the test as the behaviour of get_routes is still tested
  and therefore if other tests still run, get_routes runs as well.

### DIFF
--- a/src/webmachine_router.erl
+++ b/src/webmachine_router.erl
@@ -25,7 +25,8 @@
 -export([start_link/0,
          add_route/1,
          remove_route/1,
-         remove_resource/1]).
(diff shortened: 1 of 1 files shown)


davidcoallier:
I understand there might be reasons as why it had been set @private in the first place but that's fine, figured I'd ask with a pull-request :)

It can make it useful when automating the addition and removal of routes.

davidcoallier:
Anyone?

kellymclaughlin:
Thanks for the contribution. Merging this in.

PR #76, opened 2012-06-01
Add logging for when webmachine crashes and body exists.

cjkjellander:
We get lots of crashes when clients try to get chunked delivery, but webmachine swallows those errors. We can see the return code is 500 but nothing the error logs, since there is no logging if there already is a response body.

This commit adds logging on internal errors that has a body. Refactored out the case to a pattern
match as well.


seancribbs:
Sorry for the (inexcusably long) delay, this looks good. Merging!

PR #77, opened 2012-06-22
Incorrect accept header in some blackberry devices.

mdaguete:
Hello, 

In some blackberry devices ( problaby due locale issues)  the following incorrect accept header   appears:

"text/html,application/xhtml+xml,application/xml,application/x-javascript,_/_;q=0,5"

The "q" comes from device with a ',' instead of '.'.

Regards.


kellymclaughlin:
This also addresses bug in handling `q-values` of 0. Gonna merge it in. Thanks for the pr!

PR #81, opened 2012-08-10
remove code:clash call

beerriot:
Does anyone derive use from this call?  It's the one that's responsible for log messages like:

```
** /Users/bryan/work/pmi2/riak_pipe/.eunit/riak_pipe_sup.beam hides /Users/bryan/work/pmi2/riak/deps/riak_pipe/ebin/riak_pipe_sup.beam
```

I'd prefer to have less spam in my testing output. Anyone object?


seancribbs:
:+1: It's confusing to users of Riak who are just looking at their console log too.

rzezeski:
The eunit output for riak_kv actually looks sane now.

+1 to merge.

PR #90, opened 2012-09-08
Switch from erlang:now and timer:apply_after to better alternatives

Vagabond:


beerriot:
:+1: both loggers still work for me after the change

PR #91, opened 2012-10-03
code:priv_dir change

michaelneale:
code:priv_dir causes problems (eg when you distribute by zipping up or copying an app). 

This approach avoids that.


vinoski:
Replaced with pull request #96.

michaelneale:
Nice catch on trying the priv_dir - didn't think of that !

PR #92, opened 2012-10-03
set the port via env var

michaelneale:
WEBMACHINE_PORT matches WEBMACHINE_IP - in the skel/template (only, no core changes).


vinoski:
Replaced with pull request #96.

PR #93, opened 2012-10-18
change parameterized modules to regular modules

vinoski:
The webmachine_request module and webmachine_resource module were
parameterized modules, but the OTP team recently stated they're dropping
parameterized module syntax support while keeping the underlying "tuple
module" mechanism that allows such modules to work. Change these modules
into regular modules:
1. Keep the "tuple module" parameter so that they maintain complete
   backward compatibility; no calling code has to change.
2. Add clauses for all exported functions to allow them to be called either
   using parameterized module syntax or using normal M:F(A) syntax. For
   calling webmachine_request functions with normal syntax, this means passing
   an additional argument which is a #wm_reqstate{} record; for
   webmachine_resource, the additional argument is a #wm_resource record
   (added with this commit).

Supporting normal function invocation syntax for these modules means that
new webmachine applications need not be forced to use parameterized module
invocation syntax.

Add some new unit tests to webmachine_request, but more are needed.

### DIFF
--- a/include/wm_resource.hrl
+++ b/include/wm_resource.hrl
@@ -1,0 +1,1 @@
+-record(wm_resource, {module, modstate, modexports, trace}).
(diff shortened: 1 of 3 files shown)


kellymclaughlin:
Looks good, +1 to merge.

PR #95, opened 2012-10-23
validate project name passed to new_webmachine.sh

vinoski:
The project name passed to new_webmachine.sh must be a legal Erlang
atom. If it's not, the resulting errors can be confusing to Erlang
newcomers. Use erl to ensure the name is an atom.

Addresses pull request #65.

### DIFF
--- a/scripts/new_webmachine.sh
+++ b/scripts/new_webmachine.sh
@@ -1,10 +1,24 @@
 #!/usr/bin/env bash
 
+SCRIPT=${0##*/}
 NAME=$1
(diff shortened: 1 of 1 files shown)


seancribbs:
The script name appears in the usage string twice, otherwise this is good and works for me. :+1: to merge

vinoski:
@seancribbs: I fixed the duplicate name in the usage message, thanks for noticing that.

seancribbs:
@vinoski Looks good to me.

PR #97, opened 2012-11-08
Header Rewriting

jrwest:
`RewriteMod:rewrite/2` can either return just a rewritten raw path as a string (like before) or 2-tuple, `{Headers, RawPath}`, containing rewritten headers and the rewritten raw path. One pitfall of this implementation is any sane manipulation of the Headers data strucutre will have to be done via the `mochiweb_headers` module.

### DIFF
--- a/src/webmachine.erl
+++ b/src/webmachine.erl
@@ -37,17 +37,16 @@ stop() ->
     application:stop(webmachine).
 
 new_request(mochiweb, Request) ->
-    RawPath = case application:get_env(webmachine, rewrite_module) of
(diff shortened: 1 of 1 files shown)


kellymclaughlin:
Looks good. +1 to merge.

kellymclaughlin:
Can you also please apply this to the `wrd-content-length-with-streaming` branch.

jrwest:
patch applied to `wrd-content-length-with-streaming` as well: https://github.com/basho/webmachine/commit/664cdc01bce0c2728e1499073c753db220942fe5

PR #101, opened 2012-11-13
pass method,scheme,http vsn into rewrite

jrwest:
changed `rewrite/2` to `rewrite/5`, taking additionally the HTTP Method, scheme and HTTP version. These values cannot be rewritten (the return value of the function has not changed) but they may be used when determining the rewritten headers and path.


kellymclaughlin:
Looks good. +1 to merge.

PR #102, opened 2012-11-21
Update demo app: rebar deps, ensure inets is running, improve README

mghardy:
Updated demo app to be more intuitive for those relatively new to Erlang and Webmachine. 

Demo wouldn't compile with rebar.config webmachine dependency set to 1.7.\* - updated to 1.9.*

Fixed error due to inets not running when starting demo app.

Added a couple concrete steps to README file to test the basic app and the file resource module.

### DIFF
--- a/demo/README
+++ b/demo/README
@@ -27,9 +27,17 @@ You probably want to do one of a couple of things at this point:
 1. Start up the skeleton application:
    $ ./start.sh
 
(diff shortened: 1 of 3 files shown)


kellymclaughlin:
Thanks a lot for the submission! Gonna merge this in.

PR #104, opened 2012-12-04
Reduce request overhead

kellymclaughlin:
This pull request is composed of two changes intended to reduce the overhead of webmachine requests.
1. Only keep the request state information that is embedded into the request data record as long as necessary. This avoids the overhead of large amounts of unused and repetitive request state and request data record information being aggregately accumulated during the lifetime of a single request.
2. The dict type wastes a good deal of space and is not the most efficient type when total items is relatively small. The uses of dict are cases where the size should be relatively small so using orddict instead should have space and performance benefits.

[Benchmarking results](https://gist.github.com/4200638) using a cluster of 4 virtual machines running Ubuntu 11.10 and Riak 1.2 and using the `riak_kv_yessir_backend` show no significant change in request latencies or request throughput when using these changes to webmachine. The `yessir` backend was used to eliminate the variable of disk I/O for doing throughput and latency comparisons.

### DIFF
--- a/include/wm_reqstate.hrl
+++ b/include/wm_reqstate.hrl
@@ -1,5 +1,5 @@
 -record(wm_reqstate, {socket=undefined,
-                   metadata=dict:new(),
+                   metadata=orddict:new(),
                    range=undefined,
                    peer=undefined,
                    reqdata=undefined,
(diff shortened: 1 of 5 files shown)


seancribbs:
@kellymclaughlin This looks good, is it possible to rebase so it can be merged?

kellymclaughlin:
Sure, I'll try to do that this afternoon. Thanks!

kellymclaughlin:
Pushed a branch rebased from latest `master` [here](https://github.com/basho/webmachine/tree/klm-reduce-request-overhead-rebased)

kellymclaughlin:
Closing in favor of #152

PR #106, opened 2012-12-08
Store the dispatch_list in ETS, not application:set_env

Vagabond:
Use a protected, named table, so everyone can read it, but only
webmachine_router can write to it. The owner of the table is
webmachine_router with the heir of the process being the supervisor.

If webmachine_router is restarted, the supervisor gives ownership of the
table to the newly spawned webmachine_router process again. This means
that routing configuration is not lost if webmachine_router crashes.

The overall goal of this is to allow webmachine routes to be unregistered in an application's prep_stop callback. Calls to applicationLset_env in prep_stop go into a message loop because the application process is busy calling prep_stop().

cc @jrwest

### DIFF
--- a/src/webmachine_mochiweb.erl
+++ b/src/webmachine_mochiweb.erl
@@ -54,7 +54,7 @@ start(Options) ->
       {undefined, _} -> {?MODULE, Options5};
       {PN, O6} -> {PN, O6}
     end,
-    application_set_unless_env(webmachine, dispatch_list, DispatchList),
+    webmachine_router:init_routes(DispatchList),
     application_set_unless_env(webmachine, error_handler, ErrorHandler),
     case RewriteModule of
         undefined ->
@@ -73,7 +73,7 @@ stop() ->
 
 loop(MochiReq) ->
     Req = webmachine:new_request(mochiweb, MochiReq),
-    {ok, DispatchList} = application:get_env(webmachine, dispatch_list),
+    DispatchList = webmachine_router:get_routes(),
     Host = case host_headers(Req) of
                [H|_] -> H;
                [] -> []
--- a/src/webmachine_router.erl
+++ b/src/webmachine_router.erl
@@ -26,7 +26,8 @@
          add_route/1,
          remove_route/1,
          remove_resource/1,
-         get_routes/0]).
+         get_routes/0,
+         init_routes/1]).
 
 %% gen_server callbacks
 -export([init/1,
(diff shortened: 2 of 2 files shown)


jrwest:
ran riak against this branch and did a few basho_bench runs against it + some by hand testing. No errors, so +1

src/webmachine_router.erl
```diff
@@ -88,10 +89,40 @@ remove_resource(Resource) when is_atom(Resource) ->
...
+init_routes(DefaultRoutes) ->
+    gen_server:call(?SERVER, {init_routes, DefaultRoutes}, infinity).
+
 %% @spec start_link() -> {ok, pid()} | {error, any()}
 %% @doc Starts the webmachine_router gen_server.
 start_link() ->
-  gen_server:start_link({local, ?SERVER}, ?MODULE, [], []).
+    %% We expect to only be called from webmachine_sup
+    %%
+    %% Set up the ETS configuration table, with the heir as outselves
```
russelldb: s/outselves/ourselves

src/webmachine_mochiweb.erl
```diff
@@ -73,7 +73,7 @@ stop() ->
 
 loop(MochiReq) ->
     Req = webmachine:new_request(mochiweb, MochiReq),
-    {ok, DispatchList} = application:get_env(webmachine, dispatch_list),
+    [{dispatch_list, DispatchList}] = ets:lookup(webmachine_router, dispatch_list),
```
russelldb: For the sake of encapsulation, why not go through webmachine router to get the dispatch list?

Vagabond: I don't know... performance?  I just ported the existing behaviour.

jrwest: or perhaps similar to https://github.com/basho/riak_core/commit/40d49f9864b5528b97db6a8e54626a7fb61785a7 make the ets call in a function defined in `webmachine_router`?

EDIT: just referring to encapsulation of the ets call in module that creates the table not the fallback to the gen_server part.

Vagabond: Yes, we should probably encapsulate at least with an API if not with the gen_server itself.

russelldb: Yes. I meant something like opening get_dispatch_list.

Vagabond:
Addressed concerns.

bookshelfdave:
Tested with https://github.com/basho/riak_test/pull/128, these changes seem fine. Awaiting addition changes from @Vagabond.

Vagabond:
Changed to use a public table owned by the supervisor instead.

bookshelfdave:
riak_test replication2 passes.
all eunit tests pass.
+1

PR #108, opened 2013-01-04
Use an event handler to dispatch log events

kellymclaughlin:
Use an event handler to receive and dispatch log events to any handler that are registered. This allows for more multiple handlers to be used simultaneously and allows them to be dynamically added and removed from the application.

The available logging functionality is not affected by this change. _i.e._ There are still the built in modules for access and performance logging. The change is that they are now event handlers instead of server modules.

I have also created a branch of the repo wiki, `klm-dispatch-logging`, that has a new page with details about logging setup and the available logging modules. 

For a quick setup to test both the built-in log handlers, add the following to the _app.config_ file of an application using Webmachine:

```
{webmachine, [
              {log_handlers, [{webmachine_log_handler, ["log"]},
                              {webmachine_perf_log_handler, ["log"]}]}
             ]},
```

`"log"` may be replaced with either a relative or absolute path depending on the desired location of the log files.


kellymclaughlin:
@coderoshi This pr is finally in if you're still willing to check it out.

src/webmachine_log_handler.erl
```diff
@@ -0,0 +1,116 @@
...
+
+-record(state, {hourstamp, filename, handle}).
+
+-define(FILENAME, "access.log").
+
+%% ===================================================================
+%% gen_event callbacks
+%% ===================================================================
+
+%% @private
```
reiddraper, line 45: why `@private`? same for other exported funs below.

kellymclaughlin, line 45: I think those are just copy/paste artifacts. I can take them out, but since there is no `@doc` information for any of them it probably pretty inconsequential.

reiddraper, line 45: nbd, was more asking to see if there was something I didn't know. fine to leave them

src/webmachine_mochiweb.erl
```diff
@@ -144,3 +132,22 @@ host_headers(Req) ->
...
+get_app_env(Key) ->
+    application:get_env(webmachine, Key).
+
+%% @private
+%% @doc This function is used for cases where it may be desirable to
+%% override the value that is set in the request metadata under the
+%% `resource_module' key. An example would be a pattern where a set of
+%% resource modules shares a lot of common functionality that is
+%% contained in a single module and is used as the resource in all
+%% dispatch rules and the%% `ModOpts' are used to specify a smaller
```
coderoshi: the%% comment goof

reiddraper: what?

reiddraper:
I read through the wiki branch, looks great. Supervisor structure looks good. Nice choice of doing `sync_notify`, seems like a sane speed/backpressure tradeoff. I would consider changing `refresh` messages to be sent via `call` instead of `notify`. I'm hoping to try this branch in at least one WM app soon too.

src/webmachine_mochiweb.erl
reiddraper: oh, dope, i'm stupid

coderoshi:
The etag_test eqc is failing for me. Not yet sure if related, but just a heads up.

coderoshi:
False alarm. master fails in the same way.

kellymclaughlin:
@coderoshi [This](https://github.com/basho/webmachine/pull/109) pr is to fix that.

coderoshi:
oi. Then +1 on this.

reiddraper:
+1

PR #109, opened 2013-01-04
Repair and refactoring of the etag eqc test

kellymclaughlin:
Change the `etag_prop` property to use `"*"` as the value for the `If` header if the list of values generated for that header is empty. This simultaneously corrects a bad expectation by the property and expands the scope of the test. 

This test checks the behavior of the `If-Match` and `If-None-Match` headers. Previously if the list of values to use for the header was empty, the header would be set in the HTTP request with no value which was treated the same as if the header was not present at all. This caused the test to fail because if the `IfVals` list was empty then `CurVal` would never be a member of it and the expected HTTP response code would be the opposite of what was actually received (`204` instead of `412` for `If-Match` and the reverse for `If-None-Match`). 

This pull request changes the test behavior so that if the list of generated values for the `If` header is empty, the value for the header is set to `"*"`.  This list of generated values is also used to determine the proper expectation for the HTTP response code. For reference on the meaning of `"*"` as the value for `If-Match` and `If-None-Match` headers here is some relevant text from [rfc2616, sec 14](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html):

> The meaning of "If-Match: *" is that the method SHOULD be performed if the representation selected >by the origin server (or by a cache, possibly using the Vary mechanism, see section 14.44) exists, and >MUST NOT be performed if the representation does not exist.
> 
> The meaning of "If-None-Match: *" is that the method MUST NOT be performed if the representation     >selected by the origin server (or by a cache, possibly using the Vary mechanism, see section 14.44) >exists, and SHOULD be performed if the representation does not exist. 

Also, the eunit setup has been refactored to use a test suite and use the `eqc:testing_time` function to execute for 10 seconds as opposed to a fixed 250 iterations.

### DIFF
--- a/test/etag_test.erl
+++ b/test/etag_test.erl
@@ -41,6 +41,8 @@ unique(L) ->
 etag(Bin) ->
     integer_to_list(erlang:crc32(Bin)).
 
+etag_list([]) ->
+    "*";
 etag_list(Bins) ->
     string:join([[$", etag(B), $"] || B <- Bins], ",").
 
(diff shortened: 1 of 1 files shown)


test/etag_test.erl
```diff
@@ -59,31 +61,46 @@ http_request(Match, IfVals, NewVal, Count) ->
...
+       fun setup/0,
+       fun cleanup/1,
+       [
+        {timeout, 12,
+         ?_assert(eqc:quickcheck(eqc:testing_time(10, ?QC_OUT(etag_prop()))))}
+       ]}]}.
+
+setup() ->
     %% Setup ETS table to hold current etag value
     ets:new(?MODULE, [named_table, public]),
```
coderoshi, line 106: The only way I can get this test to work is by adding

```
application:start(inets),
```

to setup(). After that this test works like a champ.

kellymclaughlin:
@coderoshi I realized I omitted some important setup and cleanup in my refactoring. I've restored that and that should resolve the issues.

coderoshi:
That's good that my accidental error triggered your memory :)

It turns out the `inets` requirement was actually triggered by eqc trying to validate it's license. I updated my license, and the issue stopped.

Anyway, works as advertised, +1

PR #110, opened 2013-01-04
Replace static webmachine.app file with webmachine.app.src

kellymclaughlin:
Because editing module lists in `ebin/webmachine.app` is just annoying.


voluntas:
+1

reiddraper:
+1

PR #118, opened 2013-01-29
Wrd content length with streaming rebased 1

reiddraper:
Currently streaming responses of the form `{stream, streambody()}` must be exhausted in order to determine the `Content-Length`. There are good reasons for this (the body may be processed by some content-encoding), but sometimes you want to just say, "hey webmachine, I _know_ the `Content-Length` and here it is". This is useful if you have large resources and don't want the response returned with `Transfer-Encoding: chunked`. This is also useful when you want to allow HEAD requests to return a `Content-Length` without having to exhaust a large stream. This patch adds a new stream type, `known_length_stream`, whose values look like:

``` erlang
-type known_length_stream() :: {known_length_stream, ContentLength :: non_neg_integer(), streambody()}
```

While this code has been used in Riak CS for the past year or so, I'm quite open to feedback on approaches to make this more general, etc.

### DIFF
--- a/src/webmachine_decision_core.erl
+++ b/src/webmachine_decision_core.erl
@@ -633,6 +633,8 @@ encode_body(Body) ->
     case Body of
         {stream, StreamBody} ->
             {stream, make_encoder_stream(Encoder, Charsetter, StreamBody)};
+        {known_length_stream, Size, StreamBody} ->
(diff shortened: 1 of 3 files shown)


kellymclaughlin:
+1 to merge

PR #119, opened 2013-02-06
Minor Dialyzer type fixes for Riak CS

slfritchie:
This doesn't actually get rid of a warning ... though I really did try, using this spec from "typer" and then editing to add the 'test' atom as possible for the last arg:

```
%% Spec to avoid "'testing' can never match the type {'wm_reqdata'..."
-spec try_path_binding(maybe_improper_list(),[nonempty_string()],[[any(),...]],integer(),[{_,_}],0 | 1,'testing' | {'wm_reqdata',_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_}) -> {'no_dispatch_match',[[any(),...]]} | {_,_,[[any(),...]],[{_,_}],string(),string()}.
```

... but it didn't work. So I'm changing the used-by-eunit-tests-only atom to make it clear that this is an ignorable warning.

### DIFF
--- a/src/webmachine_dispatcher.erl
+++ b/src/webmachine_dispatcher.erl
@@ -207,8 +207,8 @@ try_path_binding([PathSpec|Rest], PathTokens, HostRemainder, Port, HostBindings,
             PathInfo = dict:from_list(NewBindings),
             RD1 =
                 case RD of
-                    testing ->
-                        testing;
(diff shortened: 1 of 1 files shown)


kellymclaughlin:
+1 to merge

PR #121, opened 2013-02-10
add support for the Yaws webserver

vinoski:
Modify webmachine to add support for using the Yaws webserver as an alternative to Mochiweb, with the default being Mochiweb. The changes are flexible enough to allow webmachine to be easily ported to other webservers as well.


vinoski:
Closing this to redo it more cleanly.

PR #124, opened 2013-02-14
Refine range header treatment

shino:
- Add functionality to respnd with normal (non-Range) contents even if there is Range header in requests.
- Change transfer encoding of responses to not use chunked in case of single range request

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -5,5 +5,5 @@
 
 {deps, [
         {mochiweb, "1.5.1*", {git, "git://github.com/basho/mochiweb",
-                            {tag, "1.5.1p3"}}}
+                            {branch, "1.5"}}}
         ]}.
--- a/src/webmachine_request.erl
+++ b/src/webmachine_request.erl
@@ -354,7 +354,7 @@ send_ok_response({?MODULE, ReqState}=Req) ->
     RD0 = ReqState#wm_reqstate.reqdata,
     {Range, State} = get_range(Req),
     case Range of
-        X when X =:= undefined; X =:= fail ->
+        X when X =:= undefined; X =:= fail; X =:= ignore ->
             send_response(200, Req);
         Ranges ->
             {PartList, Size} = range_parts(RD0, Ranges),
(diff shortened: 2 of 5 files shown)


src/webmachine_request.erl
```diff
@@ -626,8 +602,11 @@ parts_to_body([{Start, End, Body0}], Size, Req) ->
...
                     mochiweb_util:make_io(End),
                     "/", mochiweb_util:make_io(Size)]}],
-    Body = if is_function(Body0) ->
-                   {stream, Body0(Start, End)};
+    Body = case Body0 of
+              _ when is_function(Body0) ->
+                   {known_length_stream, End - Start + 1, Body0(Start, End)};
+              {known_length_stream, ContentSize, StreamBody} ->
+                   {known_length_stream, ContentSize, StreamBody};
               true ->
```
kellymclaughlin: Need to update the catch-all here to be `_` instead of the `true` from the `if` statement.

kellymclaughlin:
I think there is an pre-existing bug [here](https://github.com/basho/webmachine/blob/master/src/webmachine_request.erl#L562) in the final function clause of `webmachine_request:range_parts/2`. I believe the first parameter should look like this instead: `#wm_reqdata{resp_body=Body0}`. I stumbled on this trying to use a `Range` header with requests using the filesystem resource that is part of the `webmachine` demo application. If you agree this is a bug, let's go ahead and fix it as part of this pr.

rebar.config
```diff
@@ -5,5 +5,5 @@
 
 {deps, [
         {mochiweb, "1.5.1*", {git, "git://github.com/basho/mochiweb",
-                            {tag, "1.5.1p3"}}}
+                            {branch, "ss-range-fix"}}}
```
kellymclaughlin: We should probably change this to specify the `1.5` branch prior to merging.

kellymclaughlin:
Regarding [this](https://github.com/basho/webmachine/pull/124#issuecomment-13628139) comment, I found [this](https://github.com/basho/webmachine/pull/52) long opened pull request, part of which fixes this very issue.

shino:
The final function clause of `webmachine_request:range_parts/2` seems to be called from the second clause:

```
range_parts(RD=#wm_reqdata{resp_body={stream, {Hunk,Next}}}, Ranges) ->
    % for now, streamed bodies are read in full for range requests
    MRB = RD#wm_reqdata.max_recv_body,
    range_parts(read_whole_stream({Hunk,Next}, [], MRB, 0), Ranges);
```

`read_whole_stream/4` returns `binary()` .

I think this clause is not a bug, but an "internal" clause.

kellymclaughlin:
Ah yes, I see now. So it should suffice just to add the missing clause.

shino:
Passing `binary()` for single range requests is possible in current master by
setting `resp_body` to `{stream, Size, RangeData::binary()}`.
Whether this API is good or not is another issue :-)

For multi range requests, I think it's not good to pass multi range data
as a _single_ binary.

I want to hold off the decision to add another clause to accept binary,
and not to include in this PR.

src/webmachine_request.erl
shino: Terrible bug... Thank you for your good eyes!

kellymclaughlin:
We can fix the handling of non-streamed requests in a separate pr, but it does need to be fixed. The problem is illustrated by the `webmachine_demo_fs_resource` module. It reads the entire contents of a file into memory and assigns it to `resp_body` and right now the result of even a single range request is a `500 Internal Server Error` due to that function clause being missing. Since webmachine does not do anything to forbid this behavior in resource modules we should fix it. Even more so since our own demo resource is broken by it. I am happy to do a separate pull request, but I think it would fit well with these changes also.

kellymclaughlin:
Made a separate [pr](https://github.com/basho/webmachine/pull/126) to address the non-streamed `Range` header issue. +1 to merge.

shino:
merged after rebase.

PR #125, opened 2013-02-18
Guess text/css MIME type for .less files

bjnortier:
This is needed to avoid 406 errors when loading & parsing Less file in the browser

### DIFF
--- a/src/webmachine_util.erl
+++ b/src/webmachine_util.erl
@@ -78,7 +78,7 @@ guess_mime(File) ->
         ".js" ->
             "application/x-javascript";
         ".less" ->
-            "text/less";
+            "text/css";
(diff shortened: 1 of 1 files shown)


kellymclaughlin:
Can you elaborate on why `text/less` instead of `test/css`?

bjnortier:
I got confused because my <link> tag was wrong - thanks for making me look again. It should be text/css. This works with text/css:

```
<link rel="stylesheet/less" href="/foo.less" type="text/css" media="screen" />
```

using the JS less parser

kellymclaughlin:
Thanks, I'll merge this in.

bjnortier:
Thanks

PR #127, opened 2013-02-21
Allow multiple webmachine instance to run in the same app

kellymclaughlin:
Changes to allow multiple instances of webmachine to coexist within
the same erlang application. The instances can use distinct dispatch
tables and be configured to listen on different IP addresses or port
numbers. 

The current behavior is maintained as the default, but may be
overridden by setting `default_router` to `false` in the application
environment and starting named router processes from the application
supervisor prior to calling `webmachine_mochiweb:start`.

The key is to provide a distinct name for each router using `webmachine:start_link/1` and to include that name as the value for the `name` key in the `Config` passed to `webmachine_mochiweb:start/1`.

### DIFF
--- a/src/webmachine_mochiweb.erl
+++ b/src/webmachine_mochiweb.erl
@@ -18,11 +18,12 @@
 -module(webmachine_mochiweb).
 -author('Justin Sheehy <justin@basho.com>').
 -author('Andy Gross <andy@basho.com>').
--export([start/1, stop/0, loop/1]).
+-export([start/1, stop/0, loop/2]).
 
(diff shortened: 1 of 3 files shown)


reiddraper:
+1! yay!

cmeiklejohn:
This is fantastic!

reiddraper:
@seancribbs was musing about a potentially better way to achieve this, as it seems to have broken Riak master. Mentioning him in case he wants to put some thoughts on the PR here.

seancribbs:
@reiddraper Actually opened a PR for it already, #128.

PR #128, opened 2013-02-22
Bugfix for multiple routers under Riak

seancribbs:
The previous addition of multiple routers -- for the benefit of launching multiple applications -- assumed that any process calling `webmachine_mochiweb:start/1` would give its name as the name of a router process. However, in `riak_core`, the name given is:
1. A string, which is not a valid registered name.
2. Related to the interface/port on which the mochiweb listener binds, not the dispatch list.

This PR changes the behavior of multiple dispatch lists slightly such that:
1. There is only one `webmachine_router` process, but it has multiple dispatch lists in its ETS table. Since the table is public, we don't need to run multiple routers, just have a way to access the dispatch lists by name.
2. The desired dispatch list name is given as the `router_name` option, defaulting to `webmachine_router`. The `name` option given to `webmachine_mochiweb:start/1` is used ONLY as part of the listener process name, defaulting to `webmachine`.

### DIFF
--- a/src/webmachine_mochiweb.erl
+++ b/src/webmachine_mochiweb.erl
@@ -104,12 +104,17 @@ handle_get_option_result(GetOptRes, _) ->
 
 get_wm_options(Options) ->
     {DispatchList, Options1} = get_option(dispatch, Options),
-    {Name, Options2} = get_option(name, Options1),
+    {Name, Options2} =
+        case get_option(name, Options1) of
+            {undefined, Opts2} ->
+                {webmachine, Opts2};
+            NRes -> NRes
+        end,
     {RName, Options3} =
         case get_option(router_name, Options2) of
             {undefined, Opts3} ->
                 {webmachine_router, Opts3};
-            Res -> Res
+            RRes -> RRes
         end,
     {WMOptions, RestOptions} = lists:foldl(fun get_wm_option/2, {[], Options3}, ?WM_OPTIONS),
     {DispatchList, Name, RName, WMOptions, RestOptions}.
--- a/src/webmachine_mochiweb.erl
+++ b/src/webmachine_mochiweb.erl
@@ -30,11 +30,11 @@
 -define (WM_OPTION_DEFAULTS, [{error_handler, webmachine_error_handler}]).
 
 start(Options) ->
-    {DispatchList, PName, RName, WMOptions, OtherOptions} = get_wm_options(Options),
-    webmachine_router:init_routes(RName, DispatchList),
+    {DispatchList, PName, DGroup, WMOptions, OtherOptions} = get_wm_options(Options),
+    webmachine_router:init_routes(DGroup, DispatchList),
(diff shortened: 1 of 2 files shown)


bookshelfdave:
This patch allows riak_ee 684409f8bb to startup. All tests pass.

:+1: from me, but probably needs a quick review from someone who knows the ins and outs of wm better than I do.

reiddraper:
Please wait for us to touch bases re: how this affects Riak CS before merging.

buddhisthead:
Just want to add my +1 here.
riak_test _anything_ crashes on master without this branch, and runs correctly with it.
It's a must-have unless we revert the prior commit.

kellymclaughlin:
Hmm I was under the impression riak was pegged to a stable tag of
webmachine. Reverting the change is not really an option because we need it
for CS. I'll be happy to look further at this when I get a few minutes
unless somebody gets to it first.

seancribbs:
@kellymclaughlin This does not require reverting anything, just adding another option to the option-list in CS, naming the "router" you want to use for each listener.

reiddraper:
I'm happy to try this patch on CS this evening or weekend, but in the meantime please just update your Riak rebar.config to use a previous commit of webmachine.

src/webmachine_mochiweb.erl
```diff
@@ -125,12 +107,17 @@ get_wm_options(Options) ->
...
-                {webmachine_router, Opts2};
-            Res ->
-                Res
+                {webmachine, Opts2};
+            NRes -> NRes
+        end,
+    {RName, Options3} =
+        case get_option(router_name, Options2) of
+            {undefined, Opts3} ->
+                {webmachine_router, Opts3};
```
kellymclaughlin: Maybe just me, but `webmachine_router` now seems like a misleading name. Would something like `dispatch_list` or `dispatch_group` be more indicative of what this is for?

kellymclaughlin: Oops sorry, i meant `router_name` and not `webmachine_router`.

seancribbs: @kellymclaughlin Yes, good point. Will change to `dispatch_list`, barring objections.

kellymclaughlin: +1

seancribbs: Actually, scratch that (post-workout-induced fuzziness), `dispatch_group` or `routes_group` makes more sense. Thoughts about the `webmachine_router` value for the default? I made it that way for compatibility but I suspect we could just use `default`.

kellymclaughlin: I think `default` works well.

reiddraper: +1 to `default` and either `dispatch_group` or `routes_group` (though there I slight prefer the singular, `route_group`)

seancribbs:
Renames are done.

kellymclaughlin:
Looks good, I really like this as opposed to the multiple router approach. +1 to merge.

PR #132, opened 2013-03-12
Allow the response behavior of 405 errors to be changed

kellymclaughlin:


shino:
I want to discuss configuration.

There are some other status codes for those users may need to customize response body.
For example: `406 Not Acceptable`, `409 Conflict`
I think that `error_response_on_XXX` for each such code is not good.

How about the configuration like `error_hanlder_for_any_error` (sorry for my poor english, please replace with better words).
With this configuration enabled, `respond/1` will call error hander for status code >= 400.

I don't know webmachine's design policy and implementation details yet.
So please tell me If I miss the whole point :)

seancribbs:
I like @shino's idea, and pushed for something like it a while back when I refactored some of those error response functions. Does this block the CS release if it is changed more significantly?

reiddraper:
> I think that error_response_on_XXX for each such code is not good.

I agree. However, if you provide your own error handler, it should be easy to only have to override the codes you want. For example, if you want to just override `500` (likely the most common scenario now), you shouldn't have to also provide responses for every other 400+ error. If you want to only override specific codes, is it as simple as?

``` erlang
-module(my_handler).
-compile(export_all).

render_error(500, Req, Reason) ->
    %% do something
    ok;
render_error(405, Req, Reason) ->
    %% and something else
    ok;
render_error(Code, Req, Reason) ->
    Req:response_body().
```

kellymclaughlin:
My focus for this is admittedly narrow as I'm checking off test failures in a test suite. I'd be happy for it to be more general purpose. Ideally I think it would be nice if there was no configuration option needed, but if the error handler module defined a way to handle the error then that was used, but if not then the default behavior of calling `respond` could be used. Perhaps a way for the error handler module to defer handling of error codes instead of only the [current catch-all behavior](https://github.com/basho/webmachine/blob/master/src/webmachine_error_handler.erl#L82)?

seancribbs:
@kellymclaughlin Maybe it could be even simpler and just be `error_response(405, "Method Not Allowed")`. Is there any reason the additional configuration option is needed? I would only be concerned about responses that are expected not to have a body, but I don't think any of the codes >= 400 have that expectation.

kellymclaughlin:
@seancribbs yeah the simpler the better. i'm going to close this and we can work on implementing in a more useful manner.

PR #133, opened 2013-03-21
Update demo, rebar.config to webmachine, "1.10.*"

nacmartin:
Webmachine dependency updated: 1.9.\* -> 1.10.\* to keep the demo working.


seancribbs:
:+1:

Issue #135, opened 2013-03-27
./scripts/new_webmachine.sh will generate old deps config


nati:
I got following error when I run ./scripts/new_webmachine.sh with master then make it.

ERROR: Dependency dir /Users/nachiueno/work/try-webmachine/snowflake/deps/webmachine failed application validation with reason:
{version_mismatch,{"/Users/nachiueno/work/try-webmachine/snowflake/deps/webmachine/src/webmachine.app.src",
                   {expected,"1.9.*"},
                   {has,"1.10.0"}}}.

It looks like generated rebar.config has old value.

kellymclaughlin:
Fix is merged. Thanks for the report.

PR #136, opened 2013-03-29
fix template rebar.config to specify correct webmachine version

chaslemley:
Fixes this issue when attempting to follow the Quick Start guide: https://github.com/basho/webmachine/issues/135


kellymclaughlin:
Thanks for the fix.

Issue #122, opened 2013-02-12
content_types_accepted


andreineculau:
This is described in the wiki as

> This is used similarly to content_types_provided, except that it is for incoming resource representations – for example, PUT requests. Handler functions usually want to use wrq:req_body(ReqData) to access the incoming request body.

along with this extra info

> POST-processing function named as a Handler by content_types_accepted

If one runs the code though, reality is different:
`content_types_accepted` is only used by `accept_helper`, which is called only when
1) the request is a POST and `post_is_create` is true
2) the request is a PUT and `is_conflict` is false

Meaning that all POST requests are skipping the handlers. By the look of the code, webmachine assumes that the resource should call the request-content-type handler inside the `process_post` callback ?

IMO either the documentation should be updated,
either webmachine should be calling the request-content-type handler before calling `process_post`.
Can anyone concur?

sebastian:
Yes, this behaviour puzzles me too.
I would also expect the handlers provided in `content_types_accepted` to be invoked, even when `post_is_create` returns false.

robertoaloi:
Bump?

nickretallack:
Wait, seriously?  It doesn't read the PUT payload until is_conflict has already returned false?  That sounds pretty backwards to me.

tmcgilchrist:
Bump?

beerriot:
I have updated https://github.com/webmachine/webmachine/wiki/Resource-Functions to be more explicit about `content_types_accepted` only applying to POST when `post_is_create` is `true`.

PR #139, opened 2013-04-19
Run content_types_provided handlers for POST

Cy6erBr4in:
Run handlers returned by contect_types_provided/2 for POST requests as well. Before this change those handlers (if there're any) were run only for GET and HEAD requests.

This change is related to the issue #122.

### DIFF
--- a/src/webmachine_decision_core.erl
+++ b/src/webmachine_decision_core.erl
@@ -518,6 +518,7 @@ decision(v3o18) ->
     BuildBody = case method() of
         'GET' -> true;
         'HEAD' -> true;
+        'POST' -> true;
         _ -> false
     end,
     FinalBody = case BuildBody of


seancribbs:
:-1: The semantics of POST are not constrained enough to say it is related to the normal GET/HEAD/PUT flow and their responses. It would be better, in my opinion, to expose a means to negotiate the response type inside `process_post`.

Cy6erBr4in:
well, OK, I agree this isn't the best solution, but currently the situation is bizarre: if a client sets an Accept header, and you don't implement content_types_provided/2 in your module, the request most likely will fail with 406 error. On the other hand if you implement content_types_provided/2, the handlers it returns will never be used for POST requests. Do you have any ideas for workarounds?

seancribbs:
Tagging @justinsheehy

vinoski:
Taking a fresh look at this, I believe @seancribbs has it correct. The code path in question under `decision(v3o18)` is clearly intended for `GET` and `HEAD`; it specifically checks for one of those methods and if present, proceeds to deal with etag generation along with `Last-Modified` and `Expires` headers, all of which would be pretty unusual for `POST` responses. This seems to mean that the `content_types_provided` handlers are intended to be invoked for `GET` and `HEAD` requests only, though I don't believe that's documented. Also, the `decision(v3o20)` implementation indicates that `process_post` should be setting a response body, so if you've written your `content_types_provided` handlers such that they can be used for methods other than just `GET` and `HEAD`, you could just have your  `process_post` code call the right handler directly.

justinsheehy:
(sorry for missing the original long-ago tag)

I concur with @seancribbs and @vinoski on this.

@Cy6erBr4in: I agree that it's a little confusing, but I think that here we are dealing with a confusing aspect of HTTP itself.

The POST method is left intentionally so undefined that some of the constraints expected on other methods do not apply, so if you decide to handle a POST we pretty much just get out of the way. I would say that it might be equally or even more confusing to some people if we had some of those other handlers apply in that case, and thus only got halfway out of the way during a POST.

PR #140, opened 2013-04-25
allow apps to supply custom reason phrases for HTTP response codes

vinoski:
Allow applications to return:

{halt, {Code::integer(), ReasonPhrase::iolist()}}

in order to provide custom reason phrases in their responses. The original
{halt, Code} also still works.


vinoski:
Closing this due to updates needed to the branch before it's ready to merge.

PR #141, opened 2013-04-25
Custom reason phrase

vinoski:
allow apps to supply custom reason phrases for HTTP response codes

Allow applications to return:

```
{halt, {Code::integer(), ReasonPhrase::iolist()}}
```

in order to provide custom reason phrases in their responses. The original `{halt, Code}` also still works.

Also, in a second commit, add the doc directory to .gitignore.


macshonle:
+1

PR #142, opened 2013-05-02
Read body when DELETE to keep alive connection

shino:
DELETE may have body and mochiweb will close socket because it's unsafe
to keep alive with remaining bytes.

### DIFF
--- a/src/webmachine_decision_core.erl
+++ b/src/webmachine_decision_core.erl
@@ -429,7 +429,11 @@ decision(v3m16) ->
 %% DELETE enacted immediately?
 %% Also where DELETE is forced.
 decision(v3m20) ->
-    decision_test(resource_call(delete_resource), true, v3m20b, 500);
+    Result = resource_call(delete_resource),
+    %% DELETE may have body and TCP connection will be closed unless body is read.
+    %% See mochiweb_request:should_close.
+    maybe_flush_body_stream(),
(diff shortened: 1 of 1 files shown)


reiddraper:
I can confirm this fixes basho/riak_cs#521, but I have a couple questions:
1. What happens if the body is > 50MB?
2. Is that whole (potentially) 50MB body buffered in memory?
3. I don't think this would actually be useful, but this does prevent the resource author from reading a greater-than 50MB stream, as a stream body.

shino:
Thanks for comments and mumble.

>    What happens if the body is > 50MB?

Changed to use streaming.

>    I don't think this would actually be useful, but this does prevent the resource author from reading a greater-than 50MB stream, as a stream body.

Moved flushing code after calling `resource_call`.

>    Is that whole (potentially) 50MB body buffered in memory?

Also changed not to keep bytes in memory.

If I missed the points discussed in mumble (I'm afraid it's probable),
could you please tell me?

reiddraper:
(cc/ @kellymclaughlin)

I think this looks great. It fixes basho/riak_cs#521, and still works if I add a call to `wrq:req_body/1` in `delete_resource` of `riak_cs_wm_bucket_policy`. Pending Kelly taking a look too, I'm +1.

kellymclaughlin:
I agree, +1 to merge

PR #143, opened 2013-05-07
webmachine_dispatcher crashes on malformed Host header

kellymclaughlin:
When a misconfigured client sends in a Host header of the form "Host: foo.bar.baz:80:80" (or really anything not of the expected form "{host}" or "{host}:{port}" webmachine handler process crashes instead of gracefully returning an error code.

CRASH REPORT Process <0.24640.19> with 0 neighbours crashed with reason: no case clause matching ["foo.bar.com","80","80"] in webmachine_dispatcher:split_host_port/1 line 55

### DIFF
--- a/src/webmachine_dispatcher.erl
+++ b/src/webmachine_dispatcher.erl
@@ -59,7 +59,10 @@ split_host_port(HostAsString, Scheme) ->
             {split_host(HostPart), default_port(Scheme)};
         [] ->
             %% no host header
-            {[], default_port(Scheme)}
+            {[], default_port(Scheme)};
+        _ ->
+            %% Invalid host header
+            {invalid_host, default_port(Scheme)}
     end.
 
 split_host(HostAsString) ->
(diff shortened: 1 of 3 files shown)


kellymclaughlin:
The PR changes the behavior to return a `400 Bad Request` instead of `500`. Curl output looks like this:

```
11:07:06:webmachine(klm-handle-invalid-host-header) $ curl -H 'Host: localhost:80:80' localhost:8000/demo/test -v* About to connect() to localhost port 8000 (#0)
*   Trying ::1...
* Connection refused
*   Trying 127.0.0.1...
* connected
* Connected to localhost (127.0.0.1) port 8000 (#0)
> GET /demo/test HTTP/1.1
> User-Agent: curl/7.28.0
> Accept: */*
> Host: localhost:80:80
>
< HTTP/1.1 400 Bad Request
< Server: MochiWeb/1.1 WebMachine/1.10.0 (never breaks eye contact)
< Date: Tue, 07 May 2013 17:07:29 GMT
< Content-Type: text/html
< Content-Length: 157
<
* Connection #0 to host localhost left intact
<html><head><title>400 Bad Request</title></head><body><h1>Bad Request</h1>Invalid Host<p><hr><address>mochiweb+webmachine web server</address></body></html>* Closing connection #0
```

tjarvstrand:
When we try this patch it causes the webmachine_perf_log_handler to crash because neither finish_time nor end_time get set in the log data.

kellymclaughlin:
I've updated the function that calculates the time difference for the performance log handler so that it can handle `undefined` parameters and will not crash the perf log handler.

seancribbs:
@kellymclaughlin :+1:

PR #146, opened 2013-05-13
Encode the error response before sending it.

ewanmellor:
Pass the response from ErrorHandler:render_error through encode_body before
sending it.  This means that if the client has sent an appropriate
Accept-Encoding or Accept-Charset header then they will get the error
response encoded correctly.

As part of this, we have to handle the situation where a charset or encoding
hasn't yet been chose (or doesn't match one of the available options) even
though the resource specifies some.  This happens if the error response is
triggered before the encoding/charset header processing.


moonpolysoft:
+1

ekimus:
+1 for this.

kellymclaughlin:
Thanks for the PR. I merged a rebased version of this PR [here](https://github.com/basho/webmachine/pull/173).

PR #147, opened 2013-05-20
fix wrq:response_code/1 to handle custom reason phrase

vinoski:
An attempt to use webmachine tracing crashed due to the wrq:response_code/1
function not handling custom reason phrases introduced in commit
dedc5736. Add a new clause for that function to handle custom reason
phrases.

### DIFF
--- a/src/webmachine_request.erl
+++ b/src/webmachine_request.erl
@@ -693,7 +693,7 @@ stream_multipart_part_helper(Fun, Rest, CType, Boundary, Size) ->
 make_code({Code, undefined}) when is_integer(Code) ->
     make_code({Code, httpd_util:reason_phrase(Code)});
 make_code({Code, ReasonPhrase}) when is_integer(Code) ->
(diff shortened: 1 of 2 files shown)


slfritchie:
Hrm, a change by https://github.com/basho/webmachine/commit/dedc5736 in make_code() uses `|` to concatenate to `ReasonPhrase`, which would fail if someone were to make a common mistake of using a solitary binary (and confusing the `iodata()` type with `iolist()` type (which I frequently do)).  If `iodata()` is indeed the intended type, then all is fine.

</digression>

+1 to this change.

vinoski:
Good catch, Scott. The code in `make_code` actually works correctly with a solitary binary, since they're allowed at the end of an iolist, but using a comma there instead of `|` is clearer and less error-prone, plus the documentation says the custom reason phrase type is `iolist()` not `iodata()`. I've added a second commit to this branch with that change.

PR #150, opened 2013-06-04
Treat empty content-type header value the same as if it were undefined

kellymclaughlin:
Clients can send a content-type header with no value and this change adds
handling to treat that case the same as if the header were undefined. This
allows request processing to proceed instead of resulting in a 500 error
response to the user.

### DIFF
--- a/src/webmachine_decision_core.erl
+++ b/src/webmachine_decision_core.erl
@@ -589,10 +589,13 @@ decision(v3p11) ->
     end.
 
 accept_helper() ->
-    CT = case get_header_val("Content-Type") of
-             undefined -> "application/octet-stream";
-             Other -> Other
-         end,
+    accept_helper(get_header_val("Content-Type")).
+
+accept_helper(undefined) ->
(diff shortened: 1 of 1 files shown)


seancribbs:
I dislike the idea of allowing bad clients to interact, but agree this should not result in a 500. Perhaps it could be modified for a 400? Or is there a specific CS use-case?

Aside from that, it seems like a reasonable change and I like the small refactor.

kellymclaughlin:
I agree with you. I very much dislike changes that make exceptions for bad clients, but I also look at it as making webmachine more flexible to give users the ability to use it in cases where they have no control over client that are interacting and no means to correct their behavior. With this change it it still possible to return a 400 for this case by peeking at the content-type in `malformed_request`, but if it is not something your application chooses to care about the request can be processed normally. 

We do have  specific case where this occurs for CS where the official AWS ruby client does in fact send a content-type header with an empty value. :(

seancribbs:
:+1: then :smiley_cat:

PR #151, opened 2013-06-10
Removing io:format/2 calls from log file processing

djnym:
The use of io:format/2 in the webmachine_log.erl means that I end up with output which I can't control via lager. Changing to error_logger:info_msg/2 means it should route correctly.


kellymclaughlin:
:+1:

PR #137, opened 2013-04-05
Increase line coverage of webmachine unit and integration tests. The

macshonle:
goal for this first round is to increase coverage of the decision core
of webmachine. For each node on the v3 HTTP activity diagram there are
one or more tests that reach that node.

There are tests to cover 406, 412, 401, 400, 301, 307, 304, 404, 201,
409, 303, 202, 200, 201, 410, 300, 500, 501, 403, 414, 415, 413,
including Vary headers, streaming bodies, MD5 sums.

Additionally, extra logging and archiving have been added to access
data on Travis CI.

Some of the previously existing tests have been fixed, as well as a
minor bug in webmachine itself.

PR #156, opened 2013-07-10
fix R15 compatibility

evanmcc:
cc @kellymclaughlin @cmeiklejohn

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,17 @@
 ERL          ?= erl
 APP          := webmachine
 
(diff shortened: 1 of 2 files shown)


kellymclaughlin:
I tested this and it looks good. I also merged in the tests from #137 that actually exercise this code and they pass for me as well. I would like to get those more extensive tests merged in the near future and they also call `crypto:md5` so my suggestion is to move the `md5` wrappers to `webmachine_util` and export them for more broad use. 

Other than that :+1:

cmeiklejohn:
:+1: yay

PR #159, opened 2013-07-11
Use rebar to define new_hash

markusn:
A rebar.config.script checks the OTP version and defines new_hash
instead of having the Makefile doing it. This works better when running
R16 and Webmachine in itself is a dependency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,6 @@
 ERL          ?= erl
 APP          := webmachine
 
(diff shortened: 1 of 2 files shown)


evanmcc:
Thanks for this.  We did go with this strategy, in the end (I didn't know about it, otherwise I would have done so from the beginning).  That said, we decided to reverse the logic, since that's friendlier going forward (i.e. we can just rm the rebar.scripts file when R15 compat is no longer an issue).  Closing this.

PR #165, opened 2013-09-08
File upload demo

voila:
A simple file upload example, to show that everything needed to parse a multipart request is available in the webmachine_multipart module (this isn't mentioned anywhere in the wiki AFAICT). 
It uses erlydtl to generate the HTML form (hopefully, this is ok).


kellymclaughlin:
+1 3119f95

kellymclaughlin:
@borshop merge

PR #166, opened 2013-09-08
fix unit tests for Erlang/OTP R16

vinoski:
Fix two problems with the unit tests for R16:
1. Code failed to compile due to warnings caused by calls to the deprecated
   crypto:md5/1 with compiler option warnings_as_errors in effect. Use the
   existing {d,old_hash} erlc define to call crypto:hash/2 instead on Erlang
   versions newer than R15B01.
2. Tests involving POSTs and status code 303 redirects failed due to a
   missing {autodirect,false} option on httpc:request/4 calls. In older
   versions of Erlang httpc didn't redirect for such calls, but that changed
   in commit 4b9b5cc6 in the Erlang/OTP github repository, which fixed
   OTP-10765 as part of the R16B release. Also verify that the presence of the
   {autoredirect,false} option doesn't break the tests for older R15 or R14
   releases.

### DIFF
--- a/test/decision_core_test.erl
+++ b/test/decision_core_test.erl
@@ -27,6 +27,14 @@
 -define(HTML_CONTENT, "<html><body>Foo</body></html>").
 -define(TEXT_CONTENT, ?HTML_CONTENT).
 
+-ifndef(old_hash).
+md5(Bin) ->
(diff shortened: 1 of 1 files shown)


seancribbs:
Works fine on R16B01 and R15. :+1: to merge

PR #169, opened 2013-10-19
Add  Request Data API function peer_port/1  to wrq, for querying peer's port number.

anish-mp:
### Feature Added

 Added  wrq   function peer_port/1   for querying  peer's  port.

Peer's  port number is  useful when identifying  cookies  for session handling.  
### Function Description

```
  wrq:peer_port(ReqData) -> Port
  Port = integer() >= 0
  ReqData =  opaque record 
```


joedevivo:
accidentally closed

beerriot:
This PR requires significant updates to be mergeable today. It's probably easier to start over fresh. I'm going to close it with two notes for the future:

1. We don't use CHANGES files here. Please do not include them in future PRs.
2. It looked like peer-port returned the port of the connected peer, while peer returned the IP of a peer named in X-Forwarded-For (if set). These should return details about the same entity (but I know that's awkward, because people disagree about whether X-Forwarded-For is even allowed to include a port).

PR #170, opened 2013-10-30
Document append_to_resp_body

agx:
All other response body related functions use _resp_ not _response_ so
having _response_ in one place is confusing.


agx:
I just noticed that there's also response_code and set_response_code. These don't have resp_ aequivalents. I'd be happy to add them and adjust the sources if this makes sense to you. It'd be more consistent IMHO.

seancribbs:
:+1: 40afead

seancribbs:
@borshop merge

PR #161, opened 2013-07-19
Expose local socket.

cmeiklejohn:

PR #175, opened 2013-11-08
Expose the actual socket with wrq:socket

choptastic:
This adds wrq:socket/1, which can be used if an app happens to need access to the socket itself, without having to include webmachine records, and without having to message around with element/2.

Originally, I though sock/1 was what I needed, but it turns out to only return the IP address. So if there's another way to do this, that'd be great.

In my case, I'm adding a websocket layer to simple bridge for backends that don't support websockets, and in order to do so, I need to hijack the socket.  While I have it working with a combination of element/2 on the Webmachine Req record, then calling webmachine_request:socket/1 on it, I'd much rather not rely on the record element positioning hack and go with something more "sanctioned".

That said, if what I'm looking for is already available without `element/2` hacks or including the webmachine .hrl files, in my app, I'd love to hear it.  But I figured since this was so quick to implement, that I might as well put in a PR.

I will note that this patch, in its current form, does work for adding a websocket layer onto webmachine with simple bridge (which I'll be putting out publicly in the next few days for folks to kick the tires).

Thanks, and have a great one!

### DIFF
--- a/src/wrq.erl
+++ b/src/wrq.erl
@@ -21,7 +21,7 @@
          path_info/1,response_code/1,req_cookie/1,req_qs/1,req_headers/1,
          req_body/1,stream_req_body/2,resp_redirect/1,resp_headers/1,
          resp_body/1,app_root/1,path_tokens/1, host_tokens/1, port/1,
-         base_uri/1,sock/1]).
+         base_uri/1,sock/1,socket/1]).
 -export([path_info/2,get_req_header/2,do_redirect/2,fresh_resp_headers/2,
          get_resp_header/2,set_resp_header/3,set_resp_headers/2,
          set_disp_path/2,set_req_body/2,set_resp_body/2,set_response_code/2,
@@ -37,7 +37,7 @@
 
 
 create(Method,Version,RawPath,Headers) ->
-	create(Method,http,Version,RawPath,Headers).
+    create(Method,http,Version,RawPath,Headers).
 create(Method,Scheme,Version,RawPath,Headers) ->
     create(#wm_reqdata{method=Method,scheme=Scheme,version=Version,
                        raw_path=RawPath,req_headers=Headers,
@@ -85,6 +85,8 @@ version(_RD = #wm_reqdata{version=Version})
 peer(_RD = #wm_reqdata{peer=Peer}) when is_list(Peer) -> Peer.
 
 sock(_RD = #wm_reqdata{sock=Sock}) when is_list(Sock) -> Sock.
+
+socket(_RD = #wm_reqdata{wm_state=#wm_reqstate{socket=Socket}}) -> Socket.
 
 app_root(_RD = #wm_reqdata{app_root=AR}) when is_list(AR) -> AR.
 


cmeiklejohn:
Does #161 not work?

choptastic:
The return value of `wrq:sock` is a string/list of the local IP address (at least for me) (ex: "127.0.0.1").  I'm looking for the actual socket itself (the port object) to the connected client, such that gen_tcp functions can be called on it. 

Or perhaps I'm misunderstanding what your patch does. Is there a way to get the connected socket itself from the string?

choptastic:
To elaborate on what I am/was doing in simple_bridge before this patch to get the socket (which also worked for my needs, but is a bit nasty) was this:

``` erlang
socket(Req) ->
    ReqState = element(7, Req), %% get the wm_reqstate without record syntax (nasty)
    {Socket, _} = webmachine_request:socket(ReqState),  %% Get the socket itself
    Socket.
```

Can this be done using `wrq:sock/1`?

joedevivo:
accidentally closed

choptastic:
I'd still like to see this merged (or added in some other way), but I'm functional without it, just breaking apart the Req elements manually using `element/2` - and recognizing how fragile a solution that is.

As it is right now, the relevant code in simple_bridge is here: https://github.com/nitrogen/simple_bridge/blob/master/src/webmachine_bridge_modules/webmachine_simple_bridge.erl#L74-L88

If the lack of test-coverage is what's needed for this to happen, I'd be happy to add it.

src/wrq.erl
```diff
@@ -86,6 +86,12 @@ peer(_RD = #wm_reqdata{peer=Peer}) when is_list(Peer) -> Peer.
 
 sock(_RD = #wm_reqdata{sock=Sock}) when is_list(Sock) -> Sock.
 
+socket(_RD = #wm_reqdata{wm_state=ReqState0}) ->
+    Req = webmachine_request:new(ReqState0),
+    {Socket, ReqState} = Req:socket(),
+    put(tmp_reqstate, ReqState),
```
joedevivo: Why are you putting this in the process dictionary?

src/wrq.erl
```diff
@@ -86,6 +86,12 @@ peer(_RD = #wm_reqdata{peer=Peer}) when is_list(Peer) -> Peer.
 
 sock(_RD = #wm_reqdata{sock=Sock}) when is_list(Sock) -> Sock.
 
+socket(_RD = #wm_reqdata{wm_state=ReqState0}) ->
```
joedevivo: How do you feel about a double pattern match here? `#wm_reqdata{wm_state=#wm_reqstate{socket=Socket}}`

choptastic: I'm not opposed to it. I relied on `Req:socket()` just to minimize the reliance on the internal representation of `ReqState`. I _try_ to stick to the API when available to minimize cascading coupled changes. But if you feel this is safe, then, by all means.

src/wrq.erl
choptastic: Because I was putting Req through `webmachine_request:new()` and since `Req:socket()` returns both the Socket and a (potentially) updated state with `ReqState`, I was following the conventions as they're done elsewhere in `wrq`: https://github.com/webmachine/webmachine/blob/master/src/wrq.erl#L116

I figured that if there's a chance `ReqState` is updated, and there are other places in webmachine that rely on a maintained cached ReqState in the procdict, that I should keep that updated.

src/wrq.erl
choptastic: I'll try out the double-pattern match a bit later today and let you know.  If it works, then I'm happy to update it accordingly, which eliminates the process dictionary thing as well.

joedevivo: Perfect. I'll merge straight away then.

choptastic:
+1

Tested and working for my purposes with the double-pattern match and the PR has been rebased/updated.

joedevivo:
Excellent. Thanks for your patience

choptastic:
Thanks a ton!

It'll be nice to replace the hackery of relying breaking apart the record
with a much more sane `Req:socket()` call.

I appreciate you merging!

On Thu, Aug 27, 2015 at 7:09 AM, Joe DeVivo notifications@github.com
wrote:

> Excellent. Thanks for your patience
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/webmachine/webmachine/pull/175#issuecomment-135403074
> .

## 

Jesse Gumm
Owner, Sigma Star Systems
414.940.4866 || sigma-star.com || @jessegumm

PR #176, opened 2013-11-21
Make sure Reason is a string

agx:
Since 8194983c4686945850b142e3cf330c8795a859b6 when an exception
in the app occured the console had:

Bad value on output port 'efile'

and the error log remained empty. This was due to the tupe not being
converted to string.

With this applied errors correctly show up in the log.


agx:
Any comments on this one? Without this logging is really hosed for me on current webmachine git.

seancribbs:
Seems like a reasonable change. Thanks!

PR #178, opened 2013-12-05
Corrects the edoc spec for init_routes/1

varnerac:
Corrects the edoc spec for init_routes/1


varnerac:
It is indeed a duplicate. Closing.

PR #180, opened 2013-12-06
Support IPv6 address literals when parsing Host header

seth:
IPv6 address literals are enclosed in square brackets when used in a
URL [1](http://www.ietf.org/rfc/rfc2732.txt). This patch allows the host/port parsing in dispatch/4 to
handle such values.

The adjusted parsing implementation attempts to normalize ipv6
literals and adds brackets if they are missing. Some clients (including
ibrowse 4.0.1) send a Host header with a non-bracketed address. With
the normalization, the value returned by wrq:host_tokens/1 will have
brackets if it is an ipv6 literal.

[update: new patch behavior treats unbracketed ipv6 addresses as always on default port and keeps the address "whole"]

When a non-bracketed Host header containing an ipv6 address literal is
encountered, the right thing will happen if an explicit port is given
in the header value. Otherwise, the last section of the ipv6 address
will be parsed as the port. Since the client is sending an incorrect,
it seemed not worth adding additional workarounds for this case.

Finally, this patch introduces test/wm_integration_test and a minimal
resource module to verify the Host header and host tokens
behavior. ibrowse and curl are used to make HTTP requests for
testing. In R15, httpc is unable to make ipv6 requests. Curl used as
it gives the correct behavior -- the test is guarded and will be
skipped without failure if no curl executable is found on PATH.

### DIFF
--- a/src/webmachine_dispatcher.erl
+++ b/src/webmachine_dispatcher.erl
@@ -52,17 +52,77 @@ dispatch(HostAsString, PathAsString, DispatchList, RD) ->
     try_host_binding(DispatchList, Host, Port, Path, ExtraDepth, RD).
 
 split_host_port(HostAsString, Scheme) ->
-    case string:tokens(HostAsString, ":") of
-        [HostPart, PortPart] ->
-            {split_host(HostPart), list_to_integer(PortPart)};
-        [HostPart] ->
+    case parse_port_from_host(HostAsString) of
+        {HostPart, no_port} ->
             {split_host(HostPart), default_port(Scheme)};
-        [] ->
+        {HostPart, PortPart} ->
+            {split_host(HostPart), list_to_integer(PortPart)};
+        no_host ->
             %% no host header
             {[], default_port(Scheme)};
         _ ->
             %% Invalid host header
             {invalid_host, default_port(Scheme)}
+    end.
+
+parse_port_from_host([]) ->
+    no_host;
+parse_port_from_host(S) ->
+    parse_port_from_host(lists:reverse(S), []).
+
+parse_port_from_host([], Host) ->
+    {Host, no_port};
+parse_port_from_host([C|Rest] = Host0, Acc) when C == $: orelse C == $]->
+    Port = case Acc of
+               [] ->
+                   no_port;
+               P ->
+                   P
+           end,
+    RevHost = case C of
+                  $] ->
+                      Host0;
+                  _ ->
+                      Rest
+           end,
+    Host1 = lists:reverse(RevHost),
+    {Host2, Port2} = handle_unbracketed_ipv6(Host1, Port),
+    Host = Host2,
+    {Host, Port2};
+parse_port_from_host([C|Rest], Acc) ->
+    parse_port_from_host(Rest, [C|Acc]).
+
+%% Attempt naive normalization of degenerate unbracketed ipv6 literals
+%%
+%% IPv6 address literals in URLs (and therefore in the Host header)
+%% should be enclosed in square brackets. See
+%% http://www.ietf.org/rfc/rfc2732.txt. Some HTTP clients will
+%% incorrectly remove brackets from ipv6 address literals.
+%%
+%% An unbracketed ipv6 address literal is degenerate in URLs since you
+%% can't reliably parse it for a port. We "handle" this case by
+%% assuming default port and adding brackets.
+handle_unbracketed_ipv6("[" ++ _Rest = Host, Port) ->
+    {Host, Port};
+handle_unbracketed_ipv6(Host, Port) ->
+    case is_ipv6_addr(Host) of
+        false ->
+            {Host, Port};
+        true ->
+            %% we have an unbracketed ipv6 literal. We assume default
+            %% port and add brackets to normalize.  Port here is just
+            %% the last group, may be part of address.
+            {"[" ++ Host ++ ":" ++ Port ++ "]", no_port}
+    end.
+
+%% Return true if `Host' is an IPv6 address literal. Assumes port has
+%% been removed and simply detects precense of `:'.
+is_ipv6_addr(Host) ->
+    case string:chr(Host, $:) of
+        0 ->
+            false;
+        _ ->
+            true
     end.
 
 split_host(HostAsString) ->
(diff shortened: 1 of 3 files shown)


seth:
Just found a bug in this. I will force push a fix to this branch and send a comment. The parsing is not correct for ipv6 address literals when no port is provided.

seth:
I've force pushed this branch and addressed the issue. In the process I decided to take a slightly different approach to handling the degenerate case of unbracketed ipv6 address literals. The behavior is now to assume default port for such addresses (since port cannot be reliably parsed).

Also added unit tests for the address parsing.

seancribbs:
:+1: 59c8cdb

seancribbs:
There's a few dialyzer warnings on this build:

```
webmachine_dispatcher.erl:63: The variable _ can never match since previous clauses completely covered the type 'no_host' | {[any()],'no_port' | [any(),...]}
webmachine_dispatcher.erl:216: The pattern <_Dispatch, 'invalid_host', _Port, _Path, _Depth, _RD> can never match the type <_,[nonempty_string()],integer(),[nonempty_string()],0 | 1,{'wm_reqdata',_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_}>
webmachine_mochiweb.erl:60: The pattern {'error', 'invalid_host'} can never match the type {'no_dispatch_match',{[any()],integer()},[[any(),...]]} | {_,_,[[any(),...]],integer(),[[any(),...]],[{_,_}],string(),string()}
```

Also, in the eunit phase:

```
test/wm_echo_host_header.erl:23: can't find include lib "webmachine/include/webmachine.hrl"
```

Since it's in the same app, regular `-include` should be sufficient.

seth:
I can take a look at those. Should I rebase it on master first? It's been quite awhile :)

seth:
Hmm, dialyzer is quite noisy on master from what I see (R16).

I rebased this branch, fixed the warnings you noted as well as a few others. Still does not dialyze cleanly, but it is better (fixed a bunch of warnings unrelated to this patch).

See branch: ipv6-2014 on opscode remote. I can force push here, but thought I'd wait to see how you'd prefer to handle it (and whether you want the dialyzer fixes on a separate branch).

I was not able to repro the eunit include issue.

Link: https://github.com/opscode/webmachine/tree/ipv6-2014

seancribbs:
It should be clean on master and develop, @vinoski pushed updates recently that should be green. However, w.r.t. the include, you only need to use `-include_lib` from other applications.

seth:
Hmm, well. Wonder how I was running things wrongly. Because w/ R16 and
master it sure was not clean. Is master not the right branch perhaps?
Should this be on develop?

And while I understand the include concept, I'm still confused about
how/where you see a warning.

seth:
And for me, if I change the eunit test to use include, it fails -- which actually makes sense to me. The unit test is not part of the app. 

Is this a rebar or erlang version thing? blargh.

So now I'm not clear on what you want for this to get wrapped up.

```
Erlang R16B03-1 (erts-5.10.4) [source] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Eshell V5.10.4  (abort with ^G)
```

```
rebar 2.2.0 R14B04 20140304_185901 git 2.2.0-20-g6e24cd6
```

Though the included rebar is:

```
rebar 2.1.0-pre R14B04 20131015_182912 git 2.1.0-pre-166-ged88055
```

seancribbs:
@seth Thanks for your patience, that's on our unfortunately not-public builder (a point which I will raise again in the future). Yes, it might be a rebar thing :frowning:

seth:
So with R16 (as above) in a checkout of my opscode/ipv6-2014 branch I get a clean test run with `make test` and `make dialyzer` give the following:

```
==> webmachine (compile)
dialyzer --check_plt --plt /Users/seth/.riak_combo_dialyzer_plt --apps kernel stdlib sasl erts ssl tools os_mon runtime_tools crypto inets xmerl webtool snmp public_key mnesia eunit syntax_tools compiler && \
        dialyzer --add_to_plt --plt /Users/seth/.riak_combo_dialyzer_plt --output_plt /Users/seth/.riak_combo_dialyzer_plt --apps kernel stdlib sasl erts ssl tools os_mon runtime_tools crypto inets xmerl webtool snmp public_key mnesia eunit syntax_tools compiler ; test $? -ne 1
  Checking whether the PLT /Users/seth/.riak_combo_dialyzer_plt is up-to-date... yes
done (passed successfully)
 yes
done (passed successfully)
dialyzer --check_plt --plt .local_dialyzer_plt deps/*/ebin  && \
        dialyzer --add_to_plt --plt .local_dialyzer_plt --output_plt .local_dialyzer_plt deps/*/ebin ; test $? -ne 1
  Checking whether the PLT .local_dialyzer_plt is up-to-date... yes
done (passed successfully)
 yes
done (passed successfully)
==> webmachine (dialyzer)
  Checking whether the PLT /Users/seth/.riak_combo_dialyzer_plt is up-to-date... yes
  Checking whether the PLT .local_dialyzer_plt is up-to-date... yes
  Compiling some key modules to native code... done in 0m20.42s
  Proceeding with analysis...
webmachine_decision_core.erl:29: Expression produces a value of type [any()], but this value is unmatched
webmachine_dispatcher.erl:270: The pattern 'testing_ignore_dialyzer_warning_here' can never match the type {'wm_reqdata',_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_}
webmachine_mochiweb.erl:35: Expression produces a value of type ['ok'], but this value is unmatched
webmachine_resource.erl:174: The pattern [{M, F, [RD = {'wm_reqdata', _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _}, S]} | STRest] can never match the type [{atom() | tuple(),atom(),[any()] | byte(),[any()]}]
webmachine_sup.erl:47: Expression produces a value of type 'ok' | {'error','not_found' | 'simple_one_for_one'}, but this value is unmatched
webmachine_sup.erl:48: Expression produces a value of type 'ok' | {'error','not_found' | 'restarting' | 'running' | 'simple_one_for_one'}, but this value is unmatched
Unknown types:
  calendar:day/0
  calendar:hour/0
  calendar:month/0
  calendar:year/0
 done in 0m2.72s
done (warnings were emitted)
make: *** [dialyzer] Error 2
```

seth:
@seancribbs ah, ok. (and sorry for the impatience) :)

Anyhow, let me know if you want me to try these changes on a different branch or with different tools. Otherwise, I'll assume you have what you need to get these in -- possibly with a few adjustments based on your internal build stuff.

vinoski:
@seth are you sure your rebase out from master was correct? Last week I already fixed dialyzer warnings that looked just like these.

seth:
@vinoski So my changes are on top of:

```
26d04d8 (origin/master, master) merge develop to master
```

Should I be on top of the develop branch instead?

Or does that not look right?

seth:
Looking at the diff between master and develop, it looks like the changes you're talking about (and that look like similar fixes to what I did) are on develop only.

Though I took a slightly more aggressive approach for some of the fixes -- matching on ok for some of the file ops instead of `_`.

vinoski:
Oops, yes @seth, my mistake &mdash; those last few dialyzer fixes I made last week are currently on `develop` only.

Personally, I'd say yes @seth, you should be on the `develop` branch instead. @seancribbs what do you think? If you disagree, then I'll merge `develop` to `master` again so @seth can get those dialyzer fixes.

seth:
Might be interesting to review https://github.com/opscode/webmachine/commit/8bb1ef6f67165d26c00f36c747e235919d6fd2b5
and see if some of the more strict matches are ones you'd like to include.

seancribbs:
Yes, being on develop would be better, however, @borshop should have merged correctly. Maybe I should ask for a retry?

vinoski:
@seth given @seancribb's reply, I suggest rebasing to `develop`.

joedevivo:
This made it in with #236

PR #186, opened 2014-01-29
Fix handling of unparsable range headers

nparry:
Riak CS v1.4.2 returns a 416 status code for parseable but invalid Range
headers, but a 500 status when the header can't be parsed. This makes
the behavior consistent (always return the 416 status) by properly
handling the result of 'fail' from mochiweb_http:parse_range_request

### DIFF
--- a/src/webmachine_util.erl
+++ b/src/webmachine_util.erl
@@ -390,6 +390,8 @@ now_diff_milliseconds({M,S,U}, {M1,S1,U1}) ->
 parse_range(RawRange, ResourceLength) when is_list(RawRange) ->
     parse_range(mochiweb_http:parse_range_request(RawRange), ResourceLength, []).
 
+parse_range(fail, _ResourceLength, _Acc) ->
+    [];
 parse_range([], _ResourceLength, Acc) ->
     lists:reverse(Acc);
 parse_range([Spec | Rest], ResourceLength, Acc) ->
@@ -498,6 +500,16 @@ now_diff_milliseconds_test() ->
     Early2 = {9, 9, 9},
     ?assertEqual(1000, now_diff_milliseconds(Late, Early1)),
     ?assertEqual(1000001000, now_diff_milliseconds(Late, Early2)).
+
+parse_range_test() ->
+    ValidRange = "bytes=1-2",
+    InvalidRange = "bytes=2-1",
+    EmptyRange = "bytes=",
+    UnparsableRange = "bytes=foo",
+    ?assertEqual([{1,2}], parse_range(ValidRange, 10)),
+    ?assertEqual([], parse_range(InvalidRange, 10)),
+    ?assertEqual([], parse_range(EmptyRange, 10)),
+    ?assertEqual([], parse_range(UnparsableRange, 10)).
 
 -ifdef(EQC).
 


reiddraper:
Should an unparseable range result in HTTP 400, and not 416?

nparry:
Honestly, I'm not entirely sure about the technically correct result. [Reading this](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35)...

> The recipient of a byte-range- set that includes one or more
> syntactically invalid byte-range-spec values MUST ignore the header field
> that includes that byte-range- set.

...suggests that perhaps in this case the header should just be ignored. This seems to be what S3 does.

In any event, a 500 status seems incorrect :-) In theory this could [let CS decide](https://github.com/basho/riak_cs/blob/1.4.2/src/riak_cs_wm_object.erl#L218) how to handle these cases as either a 4xx or a 200.

reiddraper:
> ...suggests that perhaps in this case the header should just be ignored.
> This seems to be what S3 does.

Yeah, good catch. That's my reading of it as well.

> In any event, a 500 status seems incorrect :-)

Indeed.

> In theory this could [let CS
> decide(https://github.com/basho/riak_cs/blob/1.4.2/src/riak_cs_wm_object.erl#L218)how to handle these cases as either a 4xx or a 200.

Cool. I like that.

reiddraper:
I'm not sure whether this was happening before, but I'm getting this error compiling:

``` shell
src/webmachine_util.erl:561: unbalanced '-endif'
```

reiddraper:
Webmachine compiles OK on the parent of this commit: 4505d206db766b191d87e200141070a64f4f83b0.

nparry:
I'll take a look - being new to Erlang I probably messed something up.
`make test` was working for me but perhaps that isn't sufficient.

src/webmachine_util.erl
```diff
@@ -499,6 +501,16 @@ now_diff_milliseconds_test() ->
...
 
+parse_range_test() ->
+    ValidRange = "bytes=1-2",
+    InvalidRange = "bytes=2-1",
+    EmptyRange = "bytes=",
+    UnparsableRange = "bytes=foo",
+    ?assertEqual([{1,2}], parse_range(ValidRange, 10)),
+    ?assertEqual([], parse_range(InvalidRange, 10)),
+    ?assertEqual([], parse_range(EmptyRange, 10)),
+    ?assertEqual([], parse_range(UnparsableRange, 10)),
```
reiddraper: This last `,` should be a `.`. That's causing the compilation error.

nparry:
Thanks, not sure how I missed that!

FWIW, I was able to collect some request traces today and it turns out the Range headers causing problems for CS have extra spaces in them between sets of ranges, like...

```
bytes=1-2, 3-4, 5-6
```

reiddraper:
> FWIW, I was able to collect some request traces today and it turns out the Range headers causing problems for CS have extra spaces in them between sets of ranges, like...

Yes, this is a known issue. A 'feature' even... S3 does not support multiple ranges.

nparry:
It wasn't the multiple ranges that caused the parsing to fail - it was the extra whitespace in the header value. From what I can tell most clients don't include the whitespace, but a small percentage do.

(Sorry if we are saying the same thing :-) )

reiddraper:
Ah ok. I'll need to read up on the spec and see if that's something we should handle, but I imagine it is. If so, sounds like it might be a Mochiweb bug?

nparry:
Yea, probably around https://github.com/basho/mochiweb/blob/1.5.1p6/src/mochiweb_http.erl#L229 - I wasn't sure if the basho repo on Github would be a better place to submit a patch, or https://github.com/mochi/mochiweb

reiddraper:
Thanks!. Merging.

PR #190, opened 2014-02-13
Wip fix logger

samuelrivas:
`webmachine_perf_log_handler` would crash if when logging a non standard HTTP method. For example:

```
curl -XFOO http://localhost:8000
```

would crash the handler with `{badarg, [{erlang, atom_to_list, ["FOO"]}, ...`

This fixes it and also fixes the imprecise typing of the `wm_log_data` record

### DIFF
--- a/include/webmachine_logger.hrl
+++ b/include/webmachine_logger.hrl
@@ -1,7 +1,10 @@
+%% method will be an atom only for valid HTTP methods and a string otherwise,
+%% see erlang:decode_packet documentation
 -record(wm_log_data,
         {resource_module :: atom(),
          start_time :: tuple(),
-         method :: atom(),
+         method :: atom() | string() | binary(),
+
          headers,
          peer,
          sock,


samuelrivas:
A comment saying why this was closed without merging would've been appreciated

massemanet:
Why was this closed? Seems very legit.

joedevivo:
I'm sorry. This got automatically closed when I deleted the `develop` branch, because we're going back to using `master`. This has me worried about what else was automatically closed.

While we're on the topic of this PR, "Wip" makes me think that this isn't done. Is it ready for review?

samuelrivas:
Ok, no worries then.

Back to the PR, I honestly don't remember why I named this WIP, is from a year ago. From what I can see in master, it seems the actual bug is fixed, but the typing (less worrying) is not.

I'll take a look later on and rebased any change that is worth merging.

joedevivo:
Thanks. Just so you know, some aggressive work on typing all of webmachine is coming soon :D

samuelrivas:
So the bug was effectively already fixed, I just rebased the minor type patch.

Feel free to merge or reject at will :)

joedevivo:
Looks like an improvement, so I'll merge it in.

Thanks for the work and your patience :D

PR #191, opened 2014-02-14
replace use of now() with os:timestamp(), to remove some lock contention

djnym:
When doing some performance analysis with a lock-counting vm I notice a bit of contention based on the use of now().  os:timestamp() should work as well in the cases I changes as far as I can tell (also tested out fine in my production cluster.


reiddraper:
- Webmachine test suite passes
- Riak CS test suite passes using this branch

+1

PR #193, opened 2014-02-24
Fix rebar.config.script for Erlang/OTP 17(rc1)

srenatus:
Beginning from R17, what is returned by `erlang`s `system_info(otp_release)` no longer follows the old schema but simply returns `"17"`.  This makes the `rebar.config.script` check for the version turn _on_ `old_hash`, which really is not what we want. :)

This change works around this issue (while I'm not sure if it's the most beautiful way to do that) by checking for the string length first.

More details on the recent version number decisions: https://groups.google.com/d/msg/erlang-programming/C--8S_0bPME/1tIq4jn118gJ


vinoski:
Closing this, since this is fixed in abfe5f58. The fix is on develop and master, both of which now build with Erlang 17.0 as of commit 50c790dd for develop and commit 26d04d8c for master.

PR #195, opened 2014-03-07
fix not_modified_l17 testcase

h0ngcha0:
time specified after If-Modified-Since should be strictly bigger than
what is specified in last_modified to have 304 returned.


kellymclaughlin:
I believe the test is correct as it is currently written. The semantics of the `If-Modified-Since` header say that `304` should be returned if the document has not been modified _since_ the `Last-Modified` time. The test case as it stands tests the case where the value for the `If-Modified-Since` header is exactly the value of the `Last-Modified` header indicating the document was modified exactly at that time not since that time.

Issue #196, opened 2014-03-13
webmachine 'hangs' when using webmachine logger 


wardbekker:
I couldn't find steps to reproduce, but when after a while when using the webmachine logger, some requests seem to hang and never give any response. After such an occurence no more request are logged to the perf log or the error log. Using R16B03-1 and latest stable webmachine on latest OSX and Ubuntu server. Had no problems with 'hangs' after I removed the webmachine logger.

Like I said, couldn't find any steps to reproduce, but as it caused severe issues for me,  I thought I create an issue anyway to give you a heads up.

cobusc:
I see an issue when rotating the log file which causes the logger to deadlock and seems to block processing of new requests. I'll send a pull request for the fix (which is trivial).

Call-cycle causing the deadlock situation:
`webmachine_log:log_access/1` --> `gen_event:sync_notify(?EVENT_LOGGER, {log_access, LogData})` --> `webmachine_access_log_handler:handle_event({log_access, ...)` -->
`webmachine_log:maybe_rotate/3` --> `webmachine_log:log_close/3` --> `webmachine_log:log_info/1` --> `gen_event:sync_notify(?EVENT_LOGGER, {log_info, LogMsg})`, which will deadlock.

I have included a (partial) backtrace below for reference. As far as I can see the `webmachine_log:log_close/3` function must not make a call to `log_info/1`, but rather to `error_logger:info_msg`, similar to what the `log_open/2` function does. Or, alternatively, `gen_event:notify` should be used. My patch will use `error_logger`.

Backtrace of the `webmachine_log_event` process (of type `gen_event`):

```
Program counter: 0x00007fceca51fa18 (gen:do_call/4 + 392)
CP: 0x0000000000000000 (invalid)
arity = 0

0x00007fce1bbb9058 Return addr 0x00007fcec81916b0 (gen_event:rpc/2 + 96)
y(0)     #Ref<0.0.1.225068>
y(1)     infinity
y(2)     {sync_notify,{log_info,[webmachine_access_log_handler,": closing log file: ","log/access.log",10]}}
y(3)     <0.89.0>
y(4)     <0.89.0>
y(5)     []

0x00007fce1bbb9090 Return addr 0x00007fcec555e8b0 (webmachine_log:log_close/3 + 128)

0x00007fce1bbb9098 Return addr 0x00007fcec555efe8 (webmachine_log:maybe_rotate/3 + 472)
[...]

PR #197, opened 2014-04-03
Fix logger deadlock situation when log file is rotated

cobusc:
I see an issue when rotating the log file which causes the logger to deadlock and seems to block processing of new requests.

Call-cycle causing the deadlock situation:
`webmachine_log:log_access/1` --> `gen_event:sync_notify(?EVENT_LOGGER, {log_access, LogData})` --> `webmachine_access_log_handler:handle_event({log_access, ...)` -->
`webmachine_log:maybe_rotate/3` --> `webmachine_log:log_close/3` --> `webmachine_log:log_info/1` --> `gen_event:sync_notify(?EVENT_LOGGER, {log_info, LogMsg})`, which will deadlock.

I have included a (partial) backtrace below for reference. As far as I can see the `webmachine_log:log_close/3` function must not make a call to `log_info/1`, but rather to `error_logger:info_msg`, similar to what the `log_open/2` function does. Or, alternatively, `gen_event:notify` should be used. My patch uses `error_logger`.

Backtrace of the `webmachine_log_event` process (of type `gen_event`):

```
Program counter: 0x00007fceca51fa18 (gen:do_call/4 + 392)
CP: 0x0000000000000000 (invalid)
arity = 0

0x00007fce1bbb9058 Return addr 0x00007fcec81916b0 (gen_event:rpc/2 + 96)
y(0)     #Ref<0.0.1.225068>
y(1)     infinity
y(2)     {sync_notify,{log_info,[webmachine_access_log_handler,": closing log file: ","log/access.log",10]}}
y(3)     <0.89.0>
y(4)     <0.89.0>
y(5)     []

0x00007fce1bbb9090 Return addr 0x00007fcec555e8b0 (webmachine_log:log_close/3 + 128)

0x00007fce1bbb9098 Return addr 0x00007fcec555efe8 (webmachine_log:maybe_rotate/3 + 472)
y(0)     {file_descriptor,prim_file,{#Port<0.4273>,11}}

0x00007fce1bbb90a8 Return addr 0x00007fcec5557710 (webmachine_access_log_handler:handle_event/2 + 200)
y(0)     {2014,4,3,10}
y(1)     []
y(2)     "log/access.log"
```

### DIFF
--- a/src/webmachine_log.erl
+++ b/src/webmachine_log.erl
@@ -135,7 +135,7 @@ log_access(#wm_log_data{}=LogData) ->
 %% @doc Close a log file.
 -spec log_close(atom(), string(), file:io_device()) -> ok | {error, term()}.
 log_close(Mod, Name, FD) ->
-    log_info([Mod, ": closing log file: ", Name, $\n]),
+    error_logger:info_msg("~p: closing log file: ~s~n", [Mod, Name]),
     file:close(FD).
 
 %% @doc Notify registered log event handler of an error event.
--- a/src/webmachine_log.erl
+++ b/src/webmachine_log.erl
@@ -135,7 +135,7 @@ log_access(#wm_log_data{}=LogData) ->
 %% @doc Close a log file.
 -spec log_close(atom(), string(), file:io_device()) -> ok | {error, term()}.
 log_close(Mod, Name, FD) ->
-    log_info([atom_to_list(Mod), ": closing log file: ", Name, $\n]),
+    error_logger:info_msg("~p: closing log file: ~s~n", [Mod, Name]),
     file:close(FD).
 
 %% @doc Notify registered log event handler of an error event.


cobusc:
This pull request addresses issue #196

cobusc:
AFAICS, this is the commit that introduced the potential logger deadlock: 8194983c4686945850b142e3cf330c8795a859b6

A quick scan of the code shows that releases 1.10.4, 1.10.4p1, 1.10.4p2 and 1.10.5 are affected.

icarus75:
Seems that 1.10.6-3-g4fe8a9a is still affected by this. The log never rotates. Erlang VM memory usage starts increasing at the point in time when log rotation should have happened. The VM process ultimately gets nuked by the OS.

cobusc:
The fix is a simple one-liner. I guess not enough people have complained about it yet.

icarus75:
That would mean webmachine logging is very rarely used. If logging were to be enabled on a busy site, it will wreak havoc.

seancribbs:
@cobusc This seems reasonable. Sorry for the delay!

seancribbs:
:+1: 6f676e6

seancribbs:
@borshop merge

PR #199, opened 2014-05-06
Update meck

jwilberding:
We need to first merge #2, then rebase this. 6 tests fail for eunit, but they fail on master, as well.


jwilberding:
Sorry, wrong place for this PR.

PR #203, opened 2014-05-28
fix version check in rebar.config.script for 17.0

vinoski:
The version check in rebar.config.script was incorrect for Erlang 17.0.


reiddraper:
+1

PR #206, opened 2014-06-16
update ibrowse to 4.0.2

kuenishi:
Riak CS also uses ibrowse 4.0.2 via [erlcloud](https://github.com/basho/erlcloud) for testing. When [trying to update webmachine](https://github.com/basho/riak_cs/pull/890) it consequently depends on ibrowse 4.0.1, which conflicts against 4.0.2. This pull request updates dependency to 4.0.2, and relieves version requirement as well.

Note: this doesn't block Riak CS work but I'd be happy if this merged before CS 1.5.0 final goes out.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -8,5 +8,5 @@
 {deps,
  [{mochiweb, "1.5.1*", {git, "git://github.com/basho/mochiweb.git", {tag, "1.5.1p7"}}},
   {meck, "0.8.2", {git, "git://github.com/basho/meck.git", {tag, "0.8.2"}}},
-  {ibrowse, "4.0.1", {git, "git://github.com/cmullaparthi/ibrowse.git", {tag, "v4.0.1"}}}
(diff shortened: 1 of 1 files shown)


seancribbs:
We need to be careful about this. Transitively dependent projects depend on specific versions of ibrowse. I'm not sure why (testing?) webmachine depends on ibrowse at all.

kuenishi:
ibrowse is used in `decision_core_test:stream_content_md5/0` (actually it's not called from eunit, because it's broken), although it looks like not hard to replace with httpc. I'm also not sure why it should be ibrowse. Maybe I can try later.

I guess the only module which directly requires ibrowse in Riak is Yokozuna. Otherwise it's for test use - or even it should be in riak_test?

seancribbs:
:+1: f99a63d

seancribbs:
@borshop merge

PR #207, opened 2014-06-27
Fix timezone calculation in webmachine_log module

sinasamavati:
I was using webmachine_log:zone/0 in a project of mine. I found a bug in the timezone calculation and fixed it in my project, so I would like to share my patch.

### DIFF
--- a/src/webmachine_log.erl
+++ b/src/webmachine_log.erl
@@ -243,16 +243,19 @@ zeropad_str(NumStr, _) ->
 
 -spec zone() -> string().
 zone() ->
-    Time = erlang:universaltime(),
-    LocalTime = calendar:universal_time_to_local_time(Time),
+    {DiffH, DiffM} = zone(erlang:universaltime(), erlang:localtime()),
+    %% Ugly reformatting code to get times like +0000 and -1300
+    if DiffH < 0 ->
+            io_lib:format("-~2..0w~2..0w", [abs(DiffH), abs(DiffM)]);
+       true ->
+            io_lib:format("+~2..0w~2..0w", [DiffH, DiffM])
+    end.
+
+-spec zone(calendar:datetime(), calendar:datetime()) -> {integer(), integer()}.
+zone(UniversalTime, LocalTime) ->
     DiffSecs = calendar:datetime_to_gregorian_seconds(LocalTime) -
-        calendar:datetime_to_gregorian_seconds(Time),
-    zone((DiffSecs/3600)*100).
-
-%% Ugly reformatting code to get times like +0000 and -1300
-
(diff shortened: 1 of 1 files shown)


joedevivo:
accidentally closed

src/webmachine_log.erl
```diff
@@ -243,16 +243,19 @@ zeropad_str(NumStr, _) ->
...
-    Time = erlang:universaltime(),
-    LocalTime = calendar:universal_time_to_local_time(Time),
-    DiffSecs = calendar:datetime_to_gregorian_seconds(LocalTime) -
-        calendar:datetime_to_gregorian_seconds(Time),
-    zone((DiffSecs/3600)*100).
-
-%% Ugly reformatting code to get times like +0000 and -1300
+    {DiffH, DiffM} = zone(erlang:universaltime(), erlang:localtime()),
+    %% Ugly reformatting code to get times like +0000 and -1300
+    if DiffH < 0 ->
```
beerriot, line 248: I'm worried that this change actually introduces two bugs.

1. Comparing one call to `erlang:universaltime/0` to another call to `erlang:localtime/0` means that there is an uncontrollable amount of skew introduced by the scheduler. This is likely _way_ below the threshold of truncation, but if `calendar:universal_time_to_local_time/1` was not the source of the bug, then I'd rather not worry about the slop at all.
2. This prints the incorrect offset for negative offsets less than an hour (`DiffH` == 0, but `DiffM` < 0). Afaict, there aren't any zones like that currently, but timezones change, and the old code didn't have this issue.

beerriot (commented):
This review is completely unfair, coming 8.5 years after the PR was opened. But, if you're still watching @sinasamavati, do you remember what the bug was that you fixed? I've commented in this review about concerns I have with the new code. But, the bug you found also isn't obvious to me, so I don't know quite how to weigh the tradeoff. Thanks for sharing in either case!

beerriot:
It looks like these are the two fixes in @sinasamavati's project that are being offered for porting here:

* https://github.com/sinasamavati/leptus/commit/936c00a87d84fcc6918dfab6848fffc18a81ceaf
* https://github.com/sinasamavati/leptus/commit/3575208b5e66d18ce720b4d99a03809eaf67d058

There's no description in those commits of what the bug was or why this fixes it. The only potential related thing I found was discussion of a bug in R12B running on FreeBSD producing a `badarg` for certain timezone+dst settings: https://erlang.org/pipermail/erlang-bugs/2008-November/001077.html We're so far beyond the versions being discussed that I'd rather wait and see if someone else can reproduce and explain the bug on current versions before merging this.

PR #217, opened 2014-09-16
Handle non-atom method names

vinoski:
When a request contains a non-standard HTTP method name, Erlang and
mochiweb return it as a string rather than an atom. This caused problems in
several parts of webmachine, including logging and tracing.

Fix all parts of the code that assumed method names were always atoms. Add
new unit tests for the fixed areas.


seancribbs:
:+1: 4b8e16c

seancribbs:
Isn't webmachine under bors?

vinoski:
Good question &mdash; I'm not sure.

PR #218, opened 2014-09-17
Change reference to listening port to 8080

derek121:
The template used by scripts/new_webmachine.sh (wmskel) sets the
listening port to 8080 (in priv/templates/src/wmskel.app.src), not 8000
(as is used in the demo app).

### DIFF
--- a/README.org
+++ b/README.org
@@ -35,7 +35,7 @@ make
 ./start.sh
 #+END_SRC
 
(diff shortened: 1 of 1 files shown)


vinoski:
@borshop +1 6712884
@borshop merge

reiddraper:
@borshop merge

PR #219, opened 2014-09-17
Change to use application:ensure_all_started/1

derek121:
When running start.sh (in demo/), when mochiweb is then started, a
failure occurs due to ssl not having been started. Using
ensure_all_started/1 rather than start/1 fixes that.

### DIFF
--- a/demo/src/webmachine_demo.erl
+++ b/demo/src/webmachine_demo.erl
@@ -4,8 +4,8 @@
 -export([start/0, start_link/0, stop/0]).
 
 ensure_started(App) ->
-    case application:start(App) of
-        ok ->
+    case application:ensure_all_started(App) of
+        {ok, _} ->
             ok;
         {error, {already_started, App}} ->
             ok


vinoski:
`application:ensure_all_started/1` was added in Erlang/OTP R16B02. Using it as you're suggesting breaks anyone using an older Erlang release.

derek121:
Sounds good- will do. Thanks for the pointer, @vinoski

derek121:
@vinoski Does your suggestion yesterday about making use of/moving the manual ensure_all_started in webmachine.erl still stand? I don't see the comment anymore.

If so, by what mechanism can webmachine:start_stop_test/0 get called (which ultimately uses the current manual ensure_all_started)?

vinoski:
@derek121 hmm, not sure what happened to my original comment &mdash; here today, gone tomorrow it seems. :-(

My suggestion is to move the `ensure_all_started/2` function used in `src/webmachine.erl` test into `src/webmachine_util.erl` as a new exported function, though I would make it arity 1 and have it call an internal arity 2 version, passing the initial empty list as the second argument. Then change the test to use that new function instead, and change the demo code to use it too.

kellymclaughlin:
Closing in favor of #226

PR #130, opened 2013-03-05
adding type/spec info to wm_reqdata

kuenishi:
For better documentation, as a first try, adding type/spec of wm_reqdata for Riak CS.

PR #221, opened 2014-09-23
Adding type/spec info to wm_reqdata

seancribbs:
Rebased from #130.


seancribbs:
@vinoski This is expected to be broken, making `wrq:wm_reqdata()` opaque opened Pandora's box.

joedevivo:
@seancribbs this got closed when I deleted `develop`. Not sure if we want to pick this up where you left off or not.

seancribbs:
Keep it closed, I think it needs to be redone.

joedevivo:
yeah! it will probably get easier as the abstraction from mochiweb goes along too

PR #226, opened 2014-11-20
Fix webmachine_demo with refactored ensure_all_started function

nickelization:
We have a special version of ensure_all_started that uses the default Erlang implementation if it's available, and otherwise falls back to our own reimplemented version of the function. I made some minor changes and factored this out into the webmachine_util module.

This function is then used in webmachine_demo to fix a startup crash that was occurring due to ssl not being started.

This PR is basically an expanded, completed version of #219 using the changes suggested by @vinoski

### DIFF
--- a/demo/src/webmachine_demo.erl
+++ b/demo/src/webmachine_demo.erl
@@ -3,35 +3,22 @@
 -author('Justin Sheehy <justin@@basho.com>').
 -export([start/0, start_link/0, stop/0]).
 
-ensure_started(App) ->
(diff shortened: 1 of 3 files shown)


kellymclaughlin:
This looks good beyond what Steve mentions about the comments. Once that is fixed this should be ready to merge.

nickelization:
Just edited the commit as requested by Steve. Thanks!

kellymclaughlin:
@borshop merge

PR #228, opened 2014-12-08
Enhance rebar.config.script for dev only dependencies

kellymclaughlin:
ibrowse and meck are only used for testing and should not be imposed
as dependencies on projects using webmachine.

rebar.config.script is used to support a dev_only_deps key in
rebar.config. The mechanism is triggered by a marker file DEV_MODE
written to the top-level of the project directory. If found, the
dev_only_deps are merged into deps. The project's Makefile is modified
to create the DEV_MODE file. Those doing dev work on webmachine can
continue to run builds via standard make commands. Projects pulling in
a dep on webmachine via rebar will receive just the needed bits.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ deps/*
 ebin/*.app
 doc/*
 .local_dialyzer_plt
+/.rebar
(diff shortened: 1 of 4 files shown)


kellymclaughlin:
@borshop merge

Issue #229, opened 2014-12-09
recent rebar update breaks new_webmachine.sh


vinoski:
The rebar update in commit 83b6dd7 broke the new_webmachine.sh script:

```
$ git checkout develop
Switched to branch 'develop'
Your branch is up-to-date with 'origin/develop'.
$ ./scripts/new_webmachine.sh testproject /tmp
==> priv (create)
ERROR: Template wmskel not found.
ERROR: create failed while processing /usr/local/src/webmachine/priv: rebar_abort
```

But the same command works fine with commit a85499e, the one just prior to commit 83b6dd7.

The problem was originally reported [on StackOverflow](http://stackoverflow.com/questions/27366971/error-setting-up-webmachine).

PR #230, opened 2014-12-09
Fix #229: make sure rebar searches for templates

vinoski:
Commit 83b6dd7 updated rebar to a version that is no longer recursive by
default when searching for templates for its "create" subcommand. Add the
"-r" option to the rebar invocation in scripts/new_webmachine.sh to make
sure it finds its templates properly.


kellymclaughlin:
+1 6d8ac79

kellymclaughlin:
@borshop merge

PR #247, opened 2015-06-23
README updates: broken URLs, org status

mbbroberg:
Some URLs broke with the organizational transition. It's also important to clarify some form of leadership, which is why Sean is mentioned directly.


seancribbs:
:+1:

PR #249, opened 2015-07-20
Rebar3 support

tsloughter:
Huge set of changes, but I thought I'd open a PR even though I don't expect it to be merged as is to get a conversation started.

I removed scripts and makefiles that shouldn't be needed any more.


seancribbs:
This is awesome @tsloughter. Could you update the `.travis.yml` to use rebar3 instead of rebar? I suspect that's what's making the build fail.

joedevivo:
I was already in the middle of this work. please hold off on the merge until I can sort it all out :D 

[edit: I missed the part about 'starting the conversation'] We'll get some parts of this in for sure. I'm cleaning a bunch of cruft in my branch, including deleting shell scripts and makefiles. We'll move the "demo" app out to it's own repo as well.

seancribbs:
. @joedevivo to the rescue

tsloughter:
Updated the travis config.

My bad @joedevivo ! Our discussion the other day got me interested and I guess I got carried away :)

joedevivo:
I left the `erlang:now()` stuff out of #250. We'll do another PR for support across versions, since I broke R15 here too.

Issue #254, opened 2015-10-07
500 error when is_authorized returns {false, RD, Ctx} as described in docs


sdebnath:
The wiki (https://github.com/webmachine/webmachine/wiki/Resource-Functions#is_authorized) mentions that we should return {true, RD, Ctx} on authorized calls and {false, RD, Ctx} on calls that are not authorized. Doing so results in 500 internal server error when wrq module tries to set response headers:

http://pastebin.com/8SFyawin

bookshelfdave:
I'm not sure if it's actually checking for a boolean:
https://github.com/webmachine/webmachine/blob/6534ee70b0c5c6b0382625d6115dbb15c0992632/src/webmachine_decision_core.erl#L213

On Wed, Oct 7, 2015 at 3:53 PM, Shawn Debnath notifications@github.com
wrote:

> The wiki (
> https://github.com/webmachine/webmachine/wiki/Resource-Functions#is_authorized)
> mentions that we should return {true, RD, Ctx} on authorized calls and
> {false, RD, Ctx} on calls that are not authorized. Doing so results in 500
> internal server error when wrq module tries to set response headers:
> 
> http://pastebin.com/8SFyawin
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/webmachine/webmachine/issues/254.

## 

Dave
https://github.com/metadave/

sdebnath:
That makes sense, I can create a PR unless you want to do it.

seancribbs:
Docs are fixed. Thanks for bringing it to our attention!

PR #255, opened 2015-10-07
If resource returns {false, ...} for is_authorized, return 401 error.

sdebnath:
The wiki mentions that we should return {true, RD, Ctx} on authorized
calls and {false, RD, Ctx} on calls that are not authorized. Returning
false results in 500 internal server error when wrq module tries to
set response headers. Ensure that we are checking for false along with
true and return 401 error. Otherwise, existing behavior is maintained.

Issue #254. Much thanks to @metadave for finding the issue

### DIFF
--- a/src/webmachine_decision_core.erl
+++ b/src/webmachine_decision_core.erl
@@ -212,6 +212,7 @@ decision(v3b9b) ->
 decision(v3b8) ->
     case resource_call(is_authorized) of
         true -> d(v3b7);
+        false -> respond(401);
         {error, Reason} ->
             error_response(Reason);
         {halt, Code}  ->


seancribbs:
:-1: I'd rather `false` be an invalid return value. The HTTP spec indicates that if you're replying with 401 that you should have a challenge in the `WWW-Authenticate` header. We can't assume anything about the desired authentication method from simply `false`.

bookshelfdave:
@seancribbs the spec in the wiki docs shows:
`-spec is_authorized(rd(), state()) -> {boolean() | string() | halt(), rd(), state()}.`, 
should it be:
`-spec is_authorized(rd(), state()) -> {true | string() | halt(), rd(), state()}.`,

seancribbs:
@metadave Yes, indeed it should. _EDIT_: I fixed it.

seancribbs:
@sdebnath Unless you're going to make further modifications, can we close this?

sdebnath:
Yes

PR #257, opened 2015-10-26
remove auth_required

seliopou:
Remove what seems to be an old resource function that is not documented anywhere and is used nowhere.


seancribbs:
:+1: Thanks!

PR #261, opened 2016-02-05
Bugfix/mra/spurious 400 response

jadeallenx:
Test case to provoke a spurious 400 response from mochiweb.  A 400 response happens when a garbage message is sent to the socket pid and then a second legitimate HTTP request hits the same socket.


jadeallenx:
Sorry. Wrong repo

PR #263, opened 2016-02-22
Added X-Forwarded-For to Access Log handler to work with load balancers

admgecko:
Added X-Forwarded-For to Access Log handler to work with load balancers

### DIFF
--- a/src/webmachine_access_log_handler.erl
+++ b/src/webmachine_access_log_handler.erl
@@ -136,33 +136,38 @@ format_req(#wm_log_data{method=Method,
             undefined -> "";
             U -> U
         end,
-    case RequestTiming of
+    XForwardedFor =
+        case mochiweb_headers:get_value("X-Forwarded-For", Headers) of
+            undefined -> "";
+            X -> X
+        end,    
+case RequestTiming of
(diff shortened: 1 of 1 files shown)


beerriot:
Thank you for the patch, but I think the actual issue here might be #315. That is, if "to work with load balancers" means "expose what a load balancer is hiding by getting the IP of the originating client in the log".

When [`webmachine_mochiweb` populates `#wm_log_data.peer`](https://github.com/webmachine/webmachine/blob/master/src/webmachine_mochiweb.erl#L153-L167), it uses the value returned by `webmachine_request:get_peer/1`. [That function calls `peer_from_peername/2`](https://github.com/webmachine/webmachine/blob/master/src/webmachine_request.erl#L125), which in turn calls `x_peername/2` if the given peer looks "local", and that function ultimately [returns part of X-Forwarded-For](https://github.com/webmachine/webmachine/blob/master/src/webmachine_request.erl#L165-L178).

However, `x_peername` returns the last element in the header instead of the first. If it returned the first, then the peer printed in the log would be the originating client.

So, I'm going to close this without merging, in favor of fixing #315.

PR #264, opened 2016-03-03
Fix mochiweb dependency so we can build with Mix

seancribbs:
This resolves errors like the below when fetching dependencies with `mix deps.get`:

```
    (mix) lib/mix/rebar.ex:201: Mix.Rebar.parse_dep({:mochiweb, nil})
```

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -5,7 +5,7 @@
 
 {xref_checks, [undefined_function_calls]}.
 
-{deps, [mochiweb]}.
+{deps, [{mochiweb, "2.12.2"}]}.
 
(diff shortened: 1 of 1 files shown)


joedevivo:
+1 mix it up!

tsloughter:
Mix :disappointed:

seancribbs:
@tsloughter It's also just better to be more specific about the dependency. :stuck_out_tongue_winking_eye:

tsloughter:
@seancribbs I know https://github.com/webmachine/webmachine/blob/master/rebar.lock#L1 :)

PR #265, opened 2016-03-17
change otp version check to accommodate patched R15B01 releases

tburghart:
If the string returned by erlang:system_info(otp_release) begins with R15B01 but has any suffix the previous version check would erroneously assume that it supported APIs introduced in R15B02.


tburghart:
This is pretty obscure and unlikely to be relevant to most anyone any more, but it can bite painfully in certain regression testing scenarios.

seancribbs:
Seems fine, merging! :smile:

PR #266, opened 2016-04-07
Fix deprecation warning

lemenkov:
Replace remaining erlang:now/0 invocations.

Signed-off-by: Peter Lemenkov lemenkov@gmail.com


seancribbs:
👍 Thanks!

PR #269, opened 2016-07-14
No one uses ping. It is cruft.

seancribbs:


joedevivo:
one plus

joedevivo:
should it be removed from the diagram?

edit: nevermind, it's not even on the diagram, showing how unnecessary it even is :P

PR #270, opened 2016-08-30
Mp/multihost host headers

marcparadise:
Do not 500 if multiple hosts present in X-Forwarded-Host/Server

X-Forwarded-For-Host/Server can contain a comma-separated list of hosts.

This change recognizes that and extracts the first host from such a list
if it's used.

Reference: https://httpd.apache.org/docs/current/mod/mod_proxy.html#x-headers.


test/decision_core_test.erl
```diff
@@ -252,26 +252,6 @@ core_tests() ->
      fun head_method_not_allowed/0,
      fun non_standard_method_501/0,
      fun non_standard_method_200/0,
-     fun bad_request_b9/0,
```
seancribbs, line 255: Why were these tests removed?

marcparadise: Nice catch, they shouldn't have been.  I was narrowing down a timeout in one of the tests, looks like I messed up adding it back  - fixing now.

marcparadise:
I've updated for the above, as well as fixing the other  inadvertent change (using HTTP 1.0 in the  `nonstandard_method_501` test instead of  HTTP 1.1)

seancribbs:
👍

PR #273, opened 2016-10-14
Adding so many types

joedevivo:


src/webmachine.erl
```diff
@@ -19,6 +19,17 @@
 -author('Andy Gross <andy@basho.com>').
 -export([start/0, stop/0, new_request/2]).
 
+-type headers() :: webmachine_headers:t().
+-type response_body() :: binary()
```
seancribbs: Must it be `binary()`? I thought `iodata()` was allowed

src/wrq.erl
```diff
@@ -204,65 +239,62 @@ merge_resp_headers(Hdrs, RD=#wm_reqdata{resp_headers=RespH}) ->
...
-            end;
-        https ->
-            case Port of
-                443 -> "";
-                _ -> ":" ++ erlang:integer_to_list(Port)
-            end;
-        _ -> ":" ++ erlang:integer_to_list(Port)
-    end.
+-spec port_string(scheme(), inet:port_number()) -> string().
+port_string(http, 80) -> "";
```
seancribbs, line 294: 👏

seancribbs (changes requested):
Other than the `binary()` / `iodata()` thing (needs investigation), this looks really good. Dialyzer likes it!

src/webmachine.erl
joedevivo: You're right, I hit something else that expects `iodata()`

seancribbs (approved):
lgtm 💯

PR #274, opened 2016-10-15
Cleaner configuration management

joedevivo:
- put explicit defaults in app.src
- don't recompute Server header per request
- Pull webmachine version for Server header from application metadata

Despite the branch name, it turns out that I just did some cleanup that could lead to lager3 in the future.

### DIFF
--- a/src/webmachine_app.erl
+++ b/src/webmachine_app.erl
@@ -27,17 +27,19 @@
 
 -include("webmachine_logger.hrl").
 
+-define(QUIP, "cafe not found").
+
 %% @spec start(_Type, _StartArgs) -> ServerRet
 %% @doc application start callback for webmachine.
 start(_Type, _StartArgs) ->
     webmachine_deps:ensure(),
+
+    %% Populate dynamic defaults on load:
+    load_default_app_config(),
+
     {ok, _Pid} = SupLinkRes = webmachine_sup:start_link(),
-    Handlers = case application:get_env(webmachine, log_handlers) of
-        undefined ->
-            [];
-        {ok, Val} ->
-            Val
-    end,
+    Handlers = application:get_env(webmachine, log_handlers, []),
+
     %% handlers failing to start are handled in the handler_watcher
     _ = [supervisor:start_child(webmachine_logger_watcher_sup,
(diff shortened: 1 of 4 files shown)


src/webmachine_mochiweb.erl
```diff
@@ -121,20 +121,21 @@ new_webmachine_req(Request) ->
...
-                       Method,
-                       Scheme,
-                       Version,
-                       mochiweb_request:get(headers, Request),
-                       mochiweb_request:get(raw_path, Request));
-        undefined ->
-            {
+    {Headers, RawPath} =
+        case application:get_env(webmachine, rewrite_module) of
+            {ok, undefined} ->
```
seancribbs, line 126: Is this `{ok, undefined}` or just `undefined`?

src/webmachine_app.erl
```diff
@@ -48,3 +50,27 @@ start(_Type, _StartArgs) ->
...
+    ok.
+
+
+set_default_server_name() ->
+    {webmachine, _, Version} =
+        lists:keyfind(webmachine, 1, application:loaded_applications()),
+    ServerName =
+        lists:flatten(
+          io_lib:format(
+            "MochiWeb/1.1 WebMachine/~s (~s)",
```
seancribbs: We should probably produce the correct mochiweb version while we're at it.

src/webmachine_mochiweb.erl
seancribbs: Oh, I see that you're setting it to undefined in the app.src.

seancribbs (changes requested):
Minor nitpicks, yadda yadda

src/webmachine_app.erl
joedevivo: Pushed, although hopefully we'll need to dynamically determine whether to include MochiWeb at all here :D

src/webmachine_mochiweb.erl
joedevivo: No implied configuration! :D

seancribbs (approved):
MOCHIWEB'D!!!!

PR #282, opened 2017-01-15
Extract initial request data from service_available in wmtrace

beerriot:
As discussed in issue webmachine/webmachine#279, commit ff44051 removed
the `ping` resource function, which wmtrace relied on to get initial
request parameters. This patch shifts `wmtrace_resource` to look for
`service_available` (the next earliest function called) instead. This
also required adding `maybe_extract_request` to the `not_exported`
branch of `aggregate_trace_part`, since not every resource exports
`service_available`, like they did `ping`.


seancribbs (approved):
Thanks Bryan! 👍

PR #285, opened 2017-10-09
no Content-Length in header if response code is 1xx or 204

PrajaktaPurohit:
Do not insert Content-Length in header if response code is 1xx or 204.

REF: https://tools.ietf.org/html/rfc7230#section-3.3.3

### DIFF
--- a/src/webmachine_request.erl
+++ b/src/webmachine_request.erl
@@ -724,24 +724,27 @@ make_version({1, 0}) ->
 make_version(_) ->
     <<"HTTP/1.1 ">>.
 
+-spec update_header_with_content_length(number(), atom(), term()) -> term().
+update_header_with_content_length(Code, _Length, RD) when (Code >= 100 andalso Code < 200) orelse
+                                                           Code =:= 204 orelse
+                                                           Code =:= 304 ->
+    mochiweb_headers:make(wrq:resp_headers(RD));
+update_header_with_content_length(_Code, Length, RD) ->
+    case Length of
+        chunked ->
+            mochiweb_headers:enter(
+              "Transfer-Encoding","chunked",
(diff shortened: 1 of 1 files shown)


src/webmachine_request.erl
```diff
@@ -724,24 +724,28 @@ make_version({1, 0}) ->
 make_version(_) ->
     <<"HTTP/1.1 ">>.
 
+-spec update_header_with_content_length(number(), atom(), term()) -> number().
+update_header_with_content_length(Code, _Length, RD) when Code >= 100,
```
seancribbs: Your guard has superfluous clauses. `when Code == 204; Code == 205; Code == 304` would be sufficient (for responses that do not require `Content-Length`).

seancribbs (changes requested):
Please add a test that demonstrates the bug and that your patch fixes the problem.

src/webmachine_request.erl
seancribbs: Also, Webmachine does not reply with responses in the `100` to `199` range (e.g. `100 Continue`), those are handled by mochiweb.

markan:
@seancribbs I'd like to try and revive this PR. 

I'm picking up this from someone else, and so missing a little bit of history and context. I see that you requested tests, and I see that there are tests here: https://github.com/webmachine/webmachine/blob/d7475f219bf2e4ae0aa151733e233e5b40f26741/src/webmachine_request.erl#L948, but I'm not sure whether those were added as a response, or were insufficient and needed improvement. Are those tests sufficient?

For cleanliness I've rebased it on master. The travis test failures seem to be the same ones mentioned in #292.

seancribbs:
Thank you for the update, @markan. I have also forgotten what other tests were needed, but the code as given fits with my knowledge of the HTTP spec. Merging.

PR #286, opened 2017-11-28
Eliminate use of module_info to check whether a module exports a function

richcarl:
Calling module_info() is comparatively expensive and is best avoided.

### DIFF
--- a/include/wm_resource.hrl
+++ b/include/wm_resource.hrl
@@ -1,1 +1,1 @@
--record(wm_resource, {module, modstate, modexports, trace}).
+-record(wm_resource, {module, modstate, trace}).
--- a/src/webmachine_resource.erl
+++ b/src/webmachine_resource.erl
@@ -17,7 +17,7 @@
 -module(webmachine_resource).
 -author('Justin Sheehy <justin@basho.com>').
 -author('Andy Gross <andy@basho.com>').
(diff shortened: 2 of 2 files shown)


seancribbs (approved):
👍  This looks good to me but I would like a second opinion.

vinoski:
LGTM too.

richcarl:
We did some profiling and found that some calls to module_info() in the system were standing out a bit. This was one of them. We were on a much older webmachine version though, so I had to forward port the change to upstream.

richcarl:
Those test failures look weird. Are they expected to work in the github travis?

beerriot:
If it's good enough for gen_server, it's good enough for me. If the module_loaded/ensure_loaded step runs into the same bottleneck, that could maybe be moved to webmachine_router:init and add_route, so it's not done on every request.

+1 thanks!

richcarl:
Note that the call to erlang:module_loaded() is fast, and the code:ensure_loaded() path is taken at most once.

richcarl:
What's your take on live code loading, btw? Leave old functions like new/4 around for a while to support upgrades, or skip such niceties?

richcarl:
Does anybody fancy merging this, or do you want something changed first?

seancribbs:
I'll yolo-merge it.

PR #287, opened 2017-11-29
Make webmachine_mochiweb loop call use latest version of module

richcarl:
This makes the code a bit more resilient to live upgrades between versions in case internal api:s change.


richcarl:
Anybody got that yolo feeling?

seancribbs:
This is fine, we will fix Travis at a later time.

PR #289, opened 2018-02-27
Cater for the lowercase content-disposition headers sent in.

marjaimate:
All HTTP headers are case insensitive, this should allow users (i.e. other libraries) to send multipart forms with that header name format.


seancribbs:
This is fine in isolation but doesn't really implement case-insensitivity, it implements the lowercase version.

marjaimate:
@seancribbs fair enough that's a valid point. I did this as it was causing me trouble on this particular header. I'll take a look and see if I can do a more generic solution.

PR #294, opened 2018-07-15
Suppress stacktrace-related warnings in OTP21

### LINKED ISSUE #293: OTP 21 issue
Webmachine is not compiling on otp21 due to webmachine_decision_core.erl:36 erlang:get_stacktrace/0: deprecated;

tazjin:
In Erlang/OTP 21 warnings are thrown for using `erlang:get_stacktrace()`, which can cause builds to fail. This PR suppresses those warnings until a decision about upgrading to the new stacktrace syntax is made.

This fixes #293.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ language: erlang
 sudo: false
 install: "true" # don't let travis run get-deps
 otp_release:
+  - 21.0
   - 20.0
   - 19.1
   - 18.3
--- a/src/webmachine_decision_core.erl
+++ b/src/webmachine_decision_core.erl
@@ -24,6 +24,12 @@
 -export([handle_request/2]).
 -include("webmachine_logger.hrl").
 
+%% Suppress Erlang/OTP 21 warnings about the new method to retrieve
+%% stacktraces.
+-ifdef(OTP_RELEASE).
+-compile({nowarn_deprecated_function, [{erlang, get_stacktrace, 0}]}).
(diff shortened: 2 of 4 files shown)


tazjin:
Ah, I guess we need to introduce a pragma to keep compatibility with older versions.

tazjin:
Okay, two options:

We can `ifdef` around the function definitions using this to toggle between the versions or suppress the warning like this:

```
-ifdef(OTP_RELEASE).
-compile({nowarn_deprecated_function, [{erlang, get_stacktrace, 0}]}).
-endif.
```

What's the webmachine policy on supporting old Erlang/OTP releases? `erlang:get_stacktrace()` will be removed in OTP 23, so at that point either support for older versions needs to be dropped or the pragma-around-functions approach becomes necessary.

tazjin:
I'm updating this PR with the version that suppresses the warnings and adding Erlang/OTP 21 to the build matrix.

tazjin:
There are test failures now but they don't seem to be related to this change ... they all seem to have `{badmatch,{error,{conn_failed,{error,eaddrnotavail}}}}` in common. Ideas?

seancribbs:
@tazjin Thank you for this patch, I'm merging it. I had a more comprehensive change on my local branch but it can wait.

Also, do not worry about that test failure, it's a strange consequence of running CI on Travis that I haven't managed to eliminate yet.

PR #297, opened 2018-12-15
OTP 21.2 compatibility

sfinnie:
Creating skeleton project fails using instructions at https://github.com/Webmachine/webmachine when building the project if using the erlang 21.2 build from erlang solutions (https://www.erlang-solutions.com/resources/download.html).  Appear to be 2 errors:

* use of erlang:get_stacktrace/0 which has now been deprecated;
* use of crypto:strong_rand_bytes/0 has replaced crypto:rand_bytes/0

PR rectifies both of these, though doesn't make any provision for building correctly on previous OTP releases.  

Note: I'm an erlang novice so please treat with appropriate caution.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,6 @@ ebin
 doc/*
 .local_dialyzer_plt
 /.rebar
+/.rebar3
 _build
 .DS_Store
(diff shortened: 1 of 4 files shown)


seancribbs:
There are already fixes for OTP 21+ on master. Apologies for letting this PR sit so long.

PR #300, opened 2019-10-31
Use correct date for L17 test

eproxus:
Using `httpd_util:rfc1123_date/1` converts the date from local time to
GMT. This makes the test case fail in any time zone east of GMT (but
pass in GMT or a time zone west of GMT). To avoid this, we have to first
convert the time to GMT via that function and then convert it back to
Erlang date time. This way, we make sure we use exactly the same date in
both Last-Modified and in If-Modified-Since, honoring the original
intent of the test case.

(cc @ThomasArts)


seancribbs:
Thanks!

PR #305, opened 2020-01-31
Fix mochiweb include for rebar2

samwar:
rebar2 doesn't like the include style for rebar3, not surprising,
so changing it to a legacy format makes rebar2 and rebar3 happy.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -5,7 +5,7 @@
 
 {xref_checks, [undefined_function_calls]}.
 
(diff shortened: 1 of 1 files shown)


samwar:
@seancribbs I was doing some testing with rebar2 and found it did not like the format of the dependency include. Can you take a look at this?

seancribbs:
I don't think it's worth supporting rebar 2 anymore. Why do you need it?

samwar:
Because my project is still using it. 😭If you don't want to merge this, I can use my fork until we migrate to another build tool in 2043. 😆

beerriot:
This happened in #309, except by that time it was mochiweb 2.22.0.

PR #306, opened 2020-04-08
webmachine_mochiweb: change LoopFun to an {M,F,As} tuple

mikpe:
The options recorded in mochiweb_http refer to a local fun that may disappear through live upgrades, leading to badfun crashes.  Since mochiweb supports `{M,F,As}` tuples in addition to funs,  we can use that instead.


seancribbs (changes requested):
Good catch, but why not change the existing `loop` function instead of making a new one that reorders the arguments?

mikpe:
Changed as requested.

seancribbs (approved):

Issue #107, opened 2012-12-18
Multipart parsing barfs on Riak link walk body


engelsanchez:
Found while writing riak_test for link walking. I think it is the parser and the body returned by the link walk query is correct. Try: 

webmachine_multipart:get_all_parts(list_to_binary("\r\n--UvEgscOlsihVbv6inze7HSj0PLe\r\nContent-Type: multipart/mixed; boundary=a6y3iiUBGN7BsyUTrtKWtSGc75X\r\n\r\n--a6y3iiUBGN7BsyUTrtKWtSGc75X\r\nX-Riak-Vclock: a85hYGBgzGDKBVIcR4M2cgdcaPTNYEpkzGNlKP7rcoovCwA=\r\nLocation: /riak/test_bucket/2\r\nContent-Type: application/octet-stream\r\nLink: </riak/test_bucket/3>; riaktag=\"next\", </riak/test_bucket>; rel=\"up\"\r\nEtag: 3GS7hwS5DlD2J2W2qkSITC\r\nLast-Modified: Tue, 18 Dec 2012 14:45:07 GMT\r\n\r\nv2\r\n--a6y3iiUBGN7BsyUTrtKWtSGc75X--\r\n\r\n--UvEgscOlsihVbv6inze7HSj0PLe--\r\n"), "UvEgscOlsihVbv6inze7HSj0PLe").

It doesn't seem to like our request+body within multipart scheme

cmeiklejohn:
I ran into this same problem with parsing another multipart response -- I think the problem here is that `webmachine_multipart:get_all_parts` is expecting the leading `\r\n` to not be present, because that's actually the header/body separator.  But, someone more familiar with webmachine should be able to confirm this.

beerriot:
This was a fun, relatively shallow rabbit hole. [The leading `\r\n` from Riak](https://github.com/basho/riak_kv/blob/develop-3.0/src/riak_kv_wm_link_walker.erl#L428) is not the header/body separator. It is, in fact, the start of the first boundary. From [section 7.2.1 of RFC1341](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html#z0):

[...]

PR #317, opened 2023-01-29
Ignore empty multipart preambles

beerriot:
This is the cheap fix for #107. The leading `\r\n` from Riak is correct behavior. This change ignores it.

As noted in the first commit message, I think this probably introduces a potential problem in other valid requests, but the chance is small, and a more complete fix would be much more invasive.

Turn on whitespace ignore to see that this is just a small change plus a test. The third commit is whitespace only to fix up this file.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -1,6 +1,6 @@
 name: Build/Test
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


beerriot:
I'm going to rebase this to see the new test flow that @seancribbs just merged.

seancribbs:
@beerriot It's not running on your branch, could you add `pull-request` to the `on` section of the workflow?

beerriot:
Hmm … it did run on my branch: https://github.com/beerriot/webmachine/actions/runs/4039678620/jobs/6944632932

But it's not connected to the pull request, apparently. Yep, I'll add `pull-request`

beerriot:
✅🎉

seancribbs (approved):

Issue #172, opened 2013-11-03
Clock Skew on 'undefined' handler


zkessin:
In the case where a web request arrives and does not match any routing rule I am seeing that fairly often it leaves a result in the perf.log that has a negative time. This is often -1ms, but can be as much as -5ms. 

I assume that there is some form of clock skew going on here.

beerriot:
I think we need more information here, like OS and OTP versions. My guess is that another process on the system (like NTP) is altering the system clock during a request. `webmachine_perf_log_handler` uses `webmachine_util:now_diff_milliseconds/2` to compute these times. The times it passes are from the `#wm_log_data` fields `start_time`, `end_time`, and `finish_time`.

```erlang
    TTPD = webmachine_util:now_diff_milliseconds(EndTime, StartTime),
    TTPS = webmachine_util:now_diff_milliseconds(FinishTime, EndTime),
```

Interestingly, when no routing rule is found, the latter two fields are `undefined`. For the TTPS (end to finish) value, this will always be zero:

```erlang
now_diff_milliseconds(undefined, undefined) ->
     0;
```

For the TTPD value (start to end), a fresh `os:timestamp` value is synthesized:

```erlang
now_diff_milliseconds(undefined, T2) ->
     now_diff_milliseconds(os:timestamp(), T2);
```

I initially expected a skew from calling `erlang:now/0` too much, but that start time is also an `os:timestamp` created in `webmachine_mochiweb`.

```erlang
    InitialLogData = #wm_log_data{start_time=os:timestamp(),
```

[The docs for `os:timetamp`](https://www.erlang.org/doc/man/os.html#timestamp-0) link to [notes on OS System Time](https://www.erlang.org/doc/apps/erts/time_correction.html#OS_System_Time) that state explicitly that no limits on its adjustment forward or backward - "That is, time warps may be observed."

[...]

PR #320, opened 2023-02-02
use monotonic time to measure request timing

beerriot:
Issue #172 noted that occasionally the perf logger would log a negative request duration. This must be due to the OS time changing (for example, because NTP is adjusting it).

Erlang docs now suggest [using `erlang:monotonic_time/0` to measure elapsed time](https://www.erlang.org/doc/apps/erts/time_correction.html#how-to-work-with-the-new-api), instead of the os:timestamp/0 we had been using, to avoid this situation, which they refer to as "time warps".

This change switches all uses of `#wm_log_data.start_time`, `.end_time`, and `.finish_time` to monotonic time.

Applications that upgrade to a version of webmachine containing this patch should also first make sure their code is [time warp safe](https://www.erlang.org/doc/apps/erts/time_correction.html#Time_Warp_Safe_Code), and then enable [multi-time warp
mode](https://www.erlang.org/doc/apps/erts/time_correction.html#multi-time-warp-mode) by adding `+C multi_time_warp` to their VM args. They should also make sure that they are *not* passing `+c false` to disable time correction. Without `+C multi_time_warp` or with `+c false`, request durations may be logged as either zero or much longer than reality in the event that system time moves backward or forward. This is not a regression, only a heads up that things can be better.

### DIFF
--- a/include/webmachine_logger.hrl
+++ b/include/webmachine_logger.hrl
@@ -2,7 +2,7 @@
 %% see erlang:decode_packet documentation
 -record(wm_log_data,
         {resource_module :: atom(),
-         start_time :: tuple(),
+         start_time :: integer(), % erlang monotonic time, native units
          method :: atom() | string() | binary(),
 
          headers,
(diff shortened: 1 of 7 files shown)


beerriot:
Ah, OTP 18 didn't support `microsecond` and `millisecond` time units. Need to change to `milli_seconds` and `micro_seconds`. https://www.erlang.org/docs/18/man/erlang.html#type-time_unit

seancribbs (approved):
😎 Great stuff!

Issue #235, opened 2015-01-26
Expect 100 may be ignored by case sensitive comparison


shino:
RFC 7231 [1] says:

```
5.1.1.  Expect

   The "Expect" header field in a request indicates a certain set of
   behaviors (expectations) that need to be supported by the server in
   order to properly handle this request.  The only such expectation
   defined by this specification is 100-continue.

     Expect  = "100-continue"

   The Expect field-value is case-insensitive.
```

In contrast, webmachine treats the value as case sensitive way [2].

For example, boto requests with `Expect: 100-Continue` (note that `C` is capital),
a sequence between boto and Riak CS is as follows:

```
PUT http://s3.amazonaws.com/6b882482-5c36-4252-adf3-2a6c00c718fe/726d4433-9aa0-4ea8-9ed5-a881edda07d4 HTTP/1.1
Accept-Encoding: identity
x-amz-content-sha256: 2edc986847e209b4016e141a6dc8716d3207350f416969382d431539bf292e4a
Content-Length: 1024
Expect: 100-Continue       %% <=== HERE ===
X-Amz-Date: 20150126T064928Z
Content-MD5: yaNM/IXZgmmMasifdgcavQ==
Content-Type: application/octet-stream
Host: s3.amazonaws.com
Authorization: AWS4-HMAC-SHA256 Credential=LMA9ZODX7LR2SIBYXKWN/20150126/us-east-1/s3/aws4_request,SignedHeaders=content-length;content-md5;content-type;expect;host;user-agent;x-amz-content-sha256;x-amz-date,Signature=0f99a5a5f8588842822062eb899666ae8dcc1e2388869cbb551130e65f3eb97b
User-Agent: Boto/2.35.1 Python/2.7.6 Linux/3.13.0-44-generic
                           %% <==== NO 100 Response around here
aaaaaaaaaaaaaaa<<snip>>

HTTP/1.1 200 OK
Server: Riak CS
ETag: "c9a34cfc85d982698c6ac89f76071abd"
Date: Mon, 26 Jan 2015 06:49:28 GMT
Content-Type: text/plain
Content-Length: 0
```

[1] https://tools.ietf.org/html/rfc7231#section-5.1.1
[2] https://github.com/basho/webmachine/blob/develop/src/webmachine_request.erl#L467

beerriot:
Simple verification against webmachine/webmachine-demo:

### Successful handling:

[...]

PR #321, opened 2023-02-02
ignore case of expect header

beerriot:
100-continue, 100-Continue, 100-CONTINUE etc. are all valid ways to specify the Expect header (see issue #235 and
https://www.rfc-editor.org/rfc/rfc7231#section-5.1.1 )


beerriot:
`{error,undef,\n        [{string,equal,[\"100-cOntinue\",\"100-continue\",true],[]}`

How far back do we want to _cough_ continue to support? 😉 willfix

seancribbs (approved):

PR #323, opened 2023-02-07
Status Code Improvements

### LINKED ISSUE #242: Add status codes defined in RFC6585 [JIRA: RIAK-1676]
Just adding to `webmachine_error.erl` will work for CS, but even without this CS just works.
- https://github.com/basho/riak_cs/pull/1118
- http://tools.ietf.org/html/rfc6585

beerriot:
I started looking into #242 and fell down the error handling rabbit hole. I've broken this change into four commits, each of which is focused on a separate issue, to hopefully make it possible to review.

The first commit (7e07f59) "fixes the glitch", by providing the registered error handler with both the status code and also the status phrase, so that if a resource returns something unofficial like `{halt, {444, "Dog Is Not Blue"}}`, the HTML error rendered by `webmachine_error_handler` says "444 Dog Is Not Blue" and not "444 Internal Server Error".

The second commit (e91677a) removes the module that looked like it should have been responsible for making this happen, but was not (`webmachine_error`).

The third commit (a25012c) actually adds status phrases for the codes requested in #242, so that a resource can just return, for example, `{halt, 428}` and not `{halt, {428, "Precondition Required"}}`.

Finally, the fourth commit (3f20e0b) is purely cleanup. Hopefully it makes status code handling and error response flows a little less cluttered.

(More details in each commit message.)

I think the first commit is enough to close #242, but I recommend the second as well, to reduce future confusion. Commits three and four are nice to haves, but I'm happy to remove them if reviewers don't agree.


seancribbs (approved):
Pretty big, but it looks clean and sets us up for any future changes to the HTTP standard w.r.t. status codes.

beerriot:
Yeah, sorry about the size. It felt better to get it done than to drag it out over multiple PRs. Thanks for reviewing & approving.

Issue #326, opened 2023-02-07
Flaky decision_core_tests


beerriot:
I've seen `decision_core_tests:non_standard_method_501` fail in two different runs, passing when rerun. It failed on just one OTP version in each run, but in different OTP versions each time.

The most recent: https://github.com/webmachine/webmachine/actions/runs/4119179995
Earlier: https://github.com/beerriot/webmachine/actions/runs/4110139686/jobs/7092675310

Both times it has been on the new match I added to check the contents of the error response body. It's hard to tell what's up, because the assertion is just that the regex matched. This is just an issue to track investigation/fix.

beerriot:
Aha! The packet we received only contained the header: https://github.com/beerriot/webmachine/actions/runs/4120012430/jobs/7114311775

We'll have to `gen_tcp:recv` again to get the body.

PR #327, opened 2023-02-08
fix error handler body content test (#326)

beerriot:
While piping custom status phrases through to the error handler in #323, an extra assertion was added to a test that was going to receive an error status that `webmachine_error_handler` would render a message body for. Unfortunately, that test communicated with Webmachine via a `gen_tcp` socket directly, instead of an http client. So occasionally, it would read only part of the response, because that's all that would show up in the first packet.

This change makes the test for rendered error body its own test, and uses httpc to make sure that the body is actually read before trying to match it.


seancribbs (approved):
Yay for improving tests! 🎉

Issue #290, opened 2018-04-04
`send_response` should come before `finish_request`


zzantozz:
In [`webmachine_decision_core:finish_response()`](https://github.com/webmachine/webmachine/blob/9c9bd3a1c7b1dbb04e356314fa5cd9a49e1910c5/src/webmachine_decision_core.erl#L90), it goes like
```
    ...
    resource_call(finish_request),
    wrcall({send_response, CodeAndPhrase}),
   ...
```
In the case of a streamed response, like `{stream, Something}`, this can lead to the `finish_request` callback happening before the stream has been fully consumed, which, in turn, can break things if your stream is some kind of stateful thing, like a database cursor, which has to be shut down after the request is "finished".

Maybe I'm doing something wrong and/or don't know about some other important concept here, but shouldn't `finish_request` only be called after the request is *really* finished? Is there another way to solve this problem?

seancribbs:
`finish_request` is typically used to fixup responses (add headers, etc) before they are sent. In the case that your stream is using a resource, it should clean them up when the stream terminates, not in `finish_request`.

zzantozz:
@seancribbs How would you define "stream terminates" when the length is unknown? What if the connection is severed before fully consuming the stream? Also, what about other potential request-related resources that shouldn't be cleaned up until the body is finished streaming? Isn't it part of the http server's job to indicate that a request has been completely serviced?

seancribbs:
You don't need to worry about the client disconnecting because each connection is handled by a separate process. If your resources (processes) that need cleanup are linked to the webmachine resource/connection, teardown will be automatic. This is an OTP principle that you can take advantage of.

Remember also that you have complete access to the request state in `finish_request`! It is not uncommon to put a flag or other term in the state to signify that you are streaming and should not reclaim resources in use.

zzantozz:
> If your resources (processes) that need cleanup are linked to the webmachine resource/connection, teardown will be automatic.

[...]

PR #330, opened 2023-02-12
don't exit(normal) on errors from mochiweb_socket:send

beerriot:
This addresses both #59 and #290.

For #59, the `normal` exit would happen before logging and tracing could be done. So if a client did something like disconnect early, and the send returned `{error, enotconn}`, no evidence of that request would be recorded.

For #290, the `normal` exit could fail to clean up resources in linked processes. Mochiweb has since changed (in 3.0.0 via mochi/mochiweb@e56a4dc) its `normal` exits to `{shutdown, Error}` exits, which will terminate linked processes correctly, as long as webmachine hasn't exit-normal'd first.

This change bubbles the error from send up through the various send helpers, in an attempt to short-circuit things like stream body producers, so as not to waste resources producing bodies that can't be sent. (The exception is "writer" body producers, which have no facility to detect such errors.) This makes the exit-shutdown change in mochiweb doubly important, since a stream can't rely on running to its end to shutdown. It couldn't rely on this anyway, with webmachine's exit-normal, but it was still effective in some circumstances.

An added benefit is that logging now records the number of bytes attempted to be sent before receiving an error, instead of the full size of the stream (since `{error, closed}` was ignored and allowed the stream to be read in full).

### DIFF
--- a/src/webmachine_request.erl
+++ b/src/webmachine_request.erl
@@ -333,50 +333,59 @@ get_outheader_value(K, {?MODULE, ReqState}) ->
     {mochiweb_headers:get_value(K,
                                 wrq:resp_headers(ReqState#wm_reqstate.reqdata)), ReqState}.
 
+-type mochiweb_socket() :: gen_tcp:socket() | {ssl, gen_tcp:socket()}.
+-spec send(mochiweb_socket(), iolist() | binary()) -> ok | {error, any()}.
 send(Socket, Data) ->
-    case mochiweb_socket:send(Socket, iolist_to_binary(Data)) of
-        ok -> ok;
-        {error,closed} -> ok;
-        _ -> exit(normal)
-    end.
+    mochiweb_socket:send(Socket, iolist_to_binary(Data)).
 
+-spec send_stream_body(mochiweb_socket(),
+                       {iolist() | binary(), fun() | done}) ->
+          {ok | {error, any()}, integer()}.
 send_stream_body(Socket, X) -> send_stream_body(Socket, X, 0).
 send_stream_body(Socket, {<<>>, done}, SoFar) ->
-    send_chunk(Socket, <<>>),
-    SoFar;
+    {Result, _} = send_chunk(Socket, <<>>),
+    {Result, SoFar};
 send_stream_body(Socket, {Data, done}, SoFar) ->
-    Size = send_chunk(Socket, Data),
+    {Result, Size} = send_chunk(Socket, Data),
     send_chunk(Socket, <<>>),
-    Size + SoFar;
+    {Result, Size + SoFar};
 send_stream_body(Socket, {<<>>, Next}, SoFar) ->
     send_stream_body(Socket, Next(), SoFar);
 send_stream_body(Socket, {[], Next}, SoFar) ->
     send_stream_body(Socket, Next(), SoFar);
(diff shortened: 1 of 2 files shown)


src/webmachine_request.erl
```diff
@@ -333,50 +333,63 @@ get_outheader_value(K, {?MODULE, ReqState}) ->
...
 send_stream_body(Socket, {Data, done}, SoFar) ->
-    Size = send_chunk(Socket, Data),
-    send_chunk(Socket, <<>>),
-    Size + SoFar;
+    case send_chunk(Socket, Data) of
+        {ok, Size} ->
+            {Result, _} = send_chunk(Socket, <<>>);
+        {Result, Size} ->
+            ok
+    end,
```
seancribbs: Modern Erlang style eschews letting bindings escape case expressions, in fact I would expect the compiler to warn about this.
```suggestion
    {Result, Size} = case send_chunk(Socket, Data) of
        {ok, Size} ->
            {Result, _} = send_chunk(Socket, <<>>),
            {Result, Size};
        Sent ->
            Sent
    end,
```

src/webmachine_request.erl
```diff
@@ -413,33 +426,40 @@ send_response(CodeAndPhrase, PassedState=#wm_reqstate{reqdata=RD}, _Req) ->
...
+        case send_head(CodeAndPhrase, PassedState, Length) of
+            ok ->
+                case wrq:method(RD) of
+                    'HEAD' ->
+                        %% See PR #329 in case of merge conflict
+                        {ok, Length};
+                    _ ->
+                        send_body(Body, PassedState, Length)
+                end
+        end,
```
seancribbs: If we are going to crash here, I would prefer a badmatch over a case_clause exception.

seancribbs (commented):
Just some minor concerns, lmk what you think.

src/webmachine_request.erl
beerriot: Oof that's going to be a hard habit to break. Re-reading what I was doing here, though, I think we can just remove the `case` entirely. We don't actually care if the `<<>>` chunk send succeeded.

src/webmachine_request.erl
beerriot: Doh. Didn't mean to crash, just forgot the error clause.

beerriot (commented):
Thanks for the review! I'll fix up these oversights.

src/webmachine_request.erl
```diff
@@ -333,50 +333,59 @@ get_outheader_value(K, {?MODULE, ReqState}) ->
...
-    Size + SoFar;
+    {Result, Size + SoFar};
 send_stream_body(Socket, {<<>>, Next}, SoFar) ->
     send_stream_body(Socket, Next(), SoFar);
 send_stream_body(Socket, {[], Next}, SoFar) ->
     send_stream_body(Socket, Next(), SoFar);
 send_stream_body(Socket, {Data, Next}, SoFar) ->
-    Size = send_chunk(Socket, Data),
-    send_stream_body(Socket, Next(), Size + SoFar).
+    case send_chunk(Socket, Data) of
```
seancribbs, line 357: 👍

seancribbs (approved):
:shipit:

beerriot:
I needed to handle the merge conflict with #329 anyway, so I rebased and squashed the work into one commit here.

Issue #248, opened 2015-06-23
500 error when browser lies about Content-Length


timclassic:
IE8 has a habit of cleanly shutting down TCP connections in between the request headers and the request body of a POST when a page navigation occurs.  I see this behavior frequently when attempting to submit a POST in a window unload handler, but could theoretically happen with any page navigation.

(Chrome handles this differently; you either get the whole POST or none at all)

When this browser behavior occurs, Webmachine crashes with the following:

```
{error,{badmatch,{error,closed}},
       [{webmachine_request,recv_unchunked_body,3,
                            [{file,"src/webmachine_request.erl"},{line,489}]},
        {webmachine_request,do_recv_body,1,
                            [{file,"src/webmachine_request.erl"},{line,448}]},
        {webmachine_request,call,2,
                            [{file,"src/webmachine_request.erl"},{line,196}]},
        {wrq,req_body,1,[{file,"src/wrq.erl"},{line,118}]},
        {rest_service_util,rd_to_event_proplist,1,
                           [{file,"src/rest_service_util.erl"},{line,339}]},
        {rest_service_util,is_authorized,2,
                           [{file,"src/rest_service_util.erl"},{line,60}]},
        {webmachine_resource,resource_call,3,
                             [{file,"src/webmachine_resource.erl"},
                              {line,186}]},
        {webmachine_resource,do,3,
                             [{file,"src/webmachine_resource.erl"},
                              {line,142}]}]}
```

Is this something I should expect Webmachine to handle more gracefully?  Obviously it is a faulty HTTP request, but I don't believe it's a crashable offense, as it is just bad input from an external source.

Depending on the response to my question, I may attempt to patch this myself and submit a PR.  Is there any guidance on handling malformed requests in Webmachine?

seancribbs:
@timclassic That's sort of a question of philosophy: is it better to let it crash or to return control to your code with an error tuple? While the error reporting can be improved, I worry that NOT crashing here would introduce subtle bugs.

A point you might want to consider -- why is your resource fetching the request body inside `is_authorized/2`? Can the request be authorized with headers alone?

timclassic:
I think that (in this case) the fact that it crashes in `is_authorized/2` is just a detail.  The user happens to be authorized and we would go on to fetch the request body eventually anyway.  I will double-check this area to make sure it's not doing anything terribly weird.

Are you concerned about subtle bugs in Webmachine, or the code using Webmachine (or both)?

[...]

PR #331, opened 2023-02-14
Better behavior when client disconnects mid-body

beerriot:
Issue #248 noted that if the client disconnects in the middle of a resource reading the request body, webmachine immediately jumps to producing an error and drops a stack trace in the error log. Random disconnects are not uncommon, and also don't really point to the sort of bugs that a resource call stack would help you diagnose. So, this PR is two proposals:

First, instead of erroring out with a `badmatch` when a socket closes mid-receive, `webmachine_request` matches the error and throws a `{webmachine_recv_error, Reason}` tuple. This allows a resource that wants to handle or silence the issue to `try wrq:req_body(ReqData) catch throw:{webmachine_recv_error, Reason} -> ... end`. Meanwhile a resource that doesn't care can just let the throw bubble on past and maintain the instant-error behavior. It was possible to do this already with `catch error:{badmatch, {error, Reason}}`, but that requires knowing webmachine internal implementation details.

Second, the place that catches the error is `webmachine_request:resource_call/3`. A catch clause matching the new throw has been added there, and it reduces the error message to just the current resource call, along with filename and line number, if available, where it left off. The idea is to point to where a try-catch could be added without making it look like `{error, closed}` is a bug in the resource.

Before these two changes, disconnecting from [`webmachine_demo_fs_resource`](https://github.com/webmachine/webmachine-demo/blob/master/src/webmachine_demo_fs_resource.erl) generated an error like this:

```
[error] path=/fs/p {error,
    {error,
        {badmatch,{error,closed}},
        [{webmachine_request,recv_unchunked_body,3,
             [{file,
                  "/Users/bryan/projects/webmachine-demo/_build/default/lib/webmachine/src/webmachine_request.erl"},
              {line,524}]},
         {webmachine_request,do_recv_body,1,
             [{file,
                  "/Users/bryan/projects/webmachine-demo/_build/default/lib/webmachine/src/webmachine_request.erl"},
              {line,478}]},
         {webmachine_request,call,2,
             [{file,
                  "/Users/bryan/projects/webmachine-demo/_build/default/lib/webmachine/src/webmachine_request.erl"},
              {line,210}]},
         {wrq,req_body,1,
             [{file,
                  "/Users/bryan/projects/webmachine-demo/_build/default/lib/webmachine/src/wrq.erl"},
              {line,146}]},
         {webmachine_demo_fs_resource,accept_content,2,
             [{file,
                  "/Users/bryan/projects/webmachine-demo/src/webmachine_demo_fs_resource.erl"},
              {line,105}]},
         {webmachine_resource,resource_call,3,
             [{file,
                  "/Users/bryan/projects/webmachine-demo/_build/default/lib/webmachine/src/webmachine_resource.erl"},
              {line,206}]},
         {webmachine_resource,do,3,
             [{file,
                  "/Users/bryan/projects/webmachine-demo/_build/default/lib/webmachine/src/webmachine_resource.erl"},
              {line,151}]},
         {webmachine_decision_core,resource_call,1,
             [{file,
                  "/Users/bryan/projects/webmachine-demo/_build/default/lib/webmachine/src/webmachine_decision_core.erl"},
              {line,48}]}]}}
```

After these two changes, the same disconnect generates an error like this:

```
[error] path=/fs/p {error,
    {throw,
        {webmachine_recv_error,closed},
        {webmachine_demo_fs_resource,accept_content,2,
            [{file,
                 "/Users/bryan/projects/webmachine-demo/src/webmachine_demo_fs_resource.erl"},
             {line,105}]}}}
```

And modifying the resource like so silences the error entirely:

```diff
     end,
-    Value = wrq:req_body(ReqData1),
-    case file:write_file(FP, Value) of
-        ok ->
-            {true, wrq:set_resp_body(Value, ReqData1), Context};
-        Err ->
-            {{error, Err}, ReqData1, Context}
+    try
+        Value = wrq:req_body(ReqData1),
+        case file:write_file(FP, Value) of
+            ok ->
+                {true, wrq:set_resp_body(Value, ReqData1), Context};
+            Err ->
+                {{error, Err}, ReqData1, Context}
+        end
+    catch  throw:{webmachine_recv_error, _} ->
+             {{halt, 400}, ReqData1, Context}
     end.
```

### DIFF
--- a/src/webmachine_request.erl
+++ b/src/webmachine_request.erl
@@ -521,13 +521,23 @@ recv_stream_body(PassedState=#wm_reqstate{reqdata=RD}, MaxHunkSize) ->
 recv_unchunked_body(Socket, MaxHunk, DataLeft) ->
     case MaxHunk >= DataLeft of
         true ->
-            {ok,Data1} = mochiweb_socket:recv(Socket,DataLeft,?IDLE_TIMEOUT),
-            {Data1, done};
+            case mochiweb_socket:recv(Socket,DataLeft,?IDLE_TIMEOUT) of
+                {ok,Data1} ->
+                    {Data1, done};
+                {error, Error} ->
+                    throw({webmachine_recv_error, Error})
+            end;
         false ->
-            {ok,Data2} = mochiweb_socket:recv(Socket,MaxHunk,?IDLE_TIMEOUT),
-            {Data2,
-             fun() -> recv_unchunked_body(Socket, MaxHunk, DataLeft-MaxHunk)
-             end}
+            case mochiweb_socket:recv(Socket,MaxHunk,?IDLE_TIMEOUT) of
+                {ok,Data2} ->
+                    {Data2,
+                     fun() ->
+                             recv_unchunked_body(
+                               Socket, MaxHunk, DataLeft-MaxHunk)
+                     end};
+                {error, Error} ->
+                    throw({webmachine_recv_error, Error})
+            end
     end.
 
 recv_chunked_body(Socket, MaxHunk) ->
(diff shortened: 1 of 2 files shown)


seancribbs (approved):
🙌 :shipit: 🚀

beerriot:
🕺 alrighty then … I wonder if either of these strategies could make other parts of webmachine more friendly …

Issue #232, opened 2015-01-16
Webmachine drops http/1.1 connections on POST request failure when tracing is enabled


sdebnath:
I have a small easy test case that repros consistently. I have a small API server built using Webmachine that is responding to REST clients. The test clients are written in erlang using eunit.

**Issue**

Erlang's httpc by default uses HTTP/1.1 to open connections with the server so we are expecting connections to be persistent regardless of responses from the server. With the latest Webmachine code and Erlang R17, turns out that enabling tracing and having 2 sequential failed (401 unauthorized) POST requests results in HTTP/1.1 connections being dropped "**error:{badmatch,{error,socket_closed_remotely}}." This is not the case when 2 similar GET requests fail in the same connection.

**Repro**

Sample webmachine code with eunit test repro tarball [[Download]](http://shawn.sdf.org/webmachine_bug.tar.bz2)

In the eunit test, we are testing a few different scenarios:

1) Two POST calls expecting HTTP 401s with inets app being stopped and restarted after each call, i.e., force a new connection. 
2) Two GET calls expecting HTTP 401s without restarting inets, i.e., reuse the same connection after a failed GET attempt in the same connection.
3) Two POST calls expecting HTTP 401s without restarting inets, i.e., reuse the same connection after a failed POST attempt in the same connection.

Test 1 and 2 have a 100% success rate, unfortunately, test 3 fails 100% of the time. (If it passes for you the first run, just retry it in eunit a couple of times).

After first trying to point the finger at httpc, suggestions on erlang-questions to try HTTP/1.0 instead of HTTP/1.1 immediately identified that the client was not the issue and pointed the server either mochiweb or webmachine to be at fault. With trying to get a clean repro ready, I found out that it was in fact enabling tracing that is causing the issue. In src/test_resource.erl init/1, you can switch tracing on and off and you will see the connections not being dropped.

**Version**
Erlang/OTP 17 [erts-6.3] [source] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]
Webmachine - latest from github

**Link to erlang-questions thread**
discussion thread: http://erlang.org/pipermail/erlang-questions/2015-January/082654.html
message with the workaround: http://erlang.org/pipermail/erlang-questions/2015-January/082656.html

**Output**

```
3> c(seq_post_tests).                    
{ok,seq_post_tests}
4> eunit:test(seq_post_tests, [verbose]).
======================== EUnit ========================
[...]

Issue #168, opened 2013-09-27
Make sure request body is consumed


reiddraper:
When a resource indicates a response code without reading the body (ex. malformed request returns true), Webmachine will send a response without ever reading the body.

According to [RFC #2616](http://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html#sec8.2.3):

```
If an origin server receives a request that does not include an
        Expect request-header field with the "100-continue" expectation,
        the request includes a request body, and the server responds
        with a final status code before reading the entire request body
        from the transport connection, then the server SHOULD NOT close
        the transport connection until it has read the entire request,
        or until the client closes the connection. Otherwise, the client
        might not reliably receive the response message. However, this
        requirement is not be construed as preventing a server from
        defending itself against denial-of-service attacks, or from
        badly broken client implementations.
```

reiddraper:
My first try is [here](https://github.com/basho/webmachine/compare/master...bugfix/ensure-body-read-before-responding). But this has an unexpected problem with requests made with except 100-continue. In that case, we don't want to read the body if the resource dismisses the request just based on the headers. Webmachine should send the non 100 code, and the client should not send the body. At least I think. This text from RFC 2616 has me wondering if the server SHOULD close the connection.

```
Upon receiving a request which includes an Expect request-header
        field with the "100-continue" expectation, an origin server MUST
        either respond with 100 (Continue) status and continue to read
        from the input stream, or respond with a final status code. The
        origin server MUST NOT wait for the request body before sending
        the 100 (Continue) response. If it responds with a final status
        code, it MAY close the transport connection or it MAY continue
        to read and discard the rest of the request.  It MUST NOT
        perform the requested method if it returns a final status code.
```

PR #332, opened 2023-02-19
Consume request body or add Connection: close header

beerriot:
Another large change … sorry @seancribbs. It's 70% test code. ✅

This resolves both #168 and #232. Just before sending the response, we now check to see if there is a request body that was not - or not fully - consumed. If there is, we attempt to consume it. If we can't, we add a "Connection: closed" header. This is because if we let an unconsumed request body get back to mochiweb, it will close the socket. Closing the socket without reading the request body causes problems for some clients. So, we're doing what we can to help them.

The major change to `webmachine_request` is to stash details about the request body consumption in the process dictionary. This allows the end-of-processing machinery to tell whether the whole request body was read, and to access the "Next()" stream body function in the case that the resource stopped part way through reading the body.

A new configuration parameter is also added, to provide some protection against malicious clients sending (potentially infinitely) large request bodies. Webmachine will attempt to flush at most 64MB before closing the connection.

The best way I could figure out to test this was to directly send two requests over the same socket, and check whether they both received responses. All of the http clients can do this, but none of them expose whether they did this. So, it's a bunch of `gen_tcp` calls to make sure all the bytes went over a single socket.

Three of the new tests passed before this change:

1. `double_get`. There are no request bodies here at all. This is just a simple test to make sure that Webmachine can actually handle two requests on the same socket. I'm not sure there was an explicit test verifying this otherwise.
2. `post_then_get`. There is a request body here, but the resource fully consumes it.
3. `delete_with_body_then_get`. Code had been added to decision core specifically to consume the bodies of DELETE requests. This code has been removed in favor of this general solution.

The other tests fail before this change because unconsumed request bodies cause mochiweb to close the connection, so the second request is not processed. Except for `too_large_unauthorized_post_then_get`, which fails because it can't read the new configuration value, but also expects the second request not to be processed because the unconsumed body of the first request was too large.

### DIFF
--- a/src/webmachine.app.src
+++ b/src/webmachine.app.src
@@ -21,6 +21,9 @@
     %% module that has rewrite/5
    ,{server_name, undefined}
     %% string() for the "Server" response header
+   ,{max_flush_bytes, 67108864}
+    %% maximum number of bytes to read from a request body that was
+    %% ignored before giving up and closing the connection (64MB default)
    ]},
 
   {maintainers,["Sean Cribbs", "Joe DeVivo", "Bryan Fink",
--- a/src/webmachine_decision_core.erl
+++ b/src/webmachine_decision_core.erl
@@ -26,7 +26,7 @@
 -include("wm_compat.hrl").
 
 handle_request(Resource, ReqState) ->
-    _ = [erase(X) || X <- [decision, code, req_body, bytes_written, tmp_reqstate]],
+    _ = [erase(X) || X <- [decision, code, req_body, bytes_written, tmp_reqstate, webmachine_stream_progress, webmachine_sent_continue]],
     put(resource, Resource),
     put(reqstate, ReqState),
     try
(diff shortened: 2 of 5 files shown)


beerriot:
Off-topic, but I learned a useful tip while preparing these tests. You can run just one suite at a time, _but you must run the `webmachine` suite first_. For example, to run just the new tests:

```
rebar3 eunit -m webmachine,multi_request_connection_test
```

But if you leave the `webmachine,` out, the suite will fail in odd ways. This is true for the other suites in the `test/` directory as well. Something about the stop-start test in the `webmachine` module gets the environment set up in a way that the other tests don't do themselves.

seancribbs (approved):
This is great! 

One thing that occurred to me we could improve for the future is handling configuration. Your comment about the tests needing to be run in certain orders reminded me of how many times in recent years on Elixir projects I've had to futz with saving and restoring the application env in tests. I think it would take a huge refactoring, but ideally we would isolate configuration to a single supervisor tree, meaning you could start as many of them as you want, in any order, without conflict. The app env settings could be defaults for unspecified settings.

beerriot:
> configuration

Yes! I know there was already a huge push to get config defaults moved into `.app.src` to be overridden by `sys.config` in applications using webmachine. That has been pretty nice for the debugging I've been doing. But you're right that for testing it would be nicer to programmatically pass in config. 🤔

Thanks for reviewing!

Issue #60, opened 2012-02-28
Errors in stream body functions when using webmachine_error_handler cause re-evaluation


beerriot:
When using the `webmachine_error_handler` module as the configured `error_handler`, encountering an error during the evaulation of a stream-producing function will cause Webmachine to call the original stream function a second time.

The `webmachine_error_handler:render_error/3` function checks the result of `webmachine_request:has_response_body/0`, and then returns that value if it exists.  The problem is that by the time this is called in the situation described above, a call to `webmachine_request:set_resp_body/1` has already been made, with the stream-producer as the argument.  So, the "body" that is now to be sent as the error response is, in fact, the "body" that just produced the error.

It's also a little bit of a problem that an error response is sent at all in this case, since it breaks the chunked encoding that was already being sent, but one thing at a time…

As an example, take the following resource, which produces one byte per second for 3 seconds, then raises an `undef` error:

``` erlang

-module(logdis_resource).
-export([init/1, to_html/2]).

-include_lib("webmachine/include/webmachine.hrl").

init([]) -> {ok, undefined}.

to_html(ReqData, State) ->
    {{stream, {<<"a">>, fun() -> slow_response(10) end}}, ReqData, State}.

slow_response(0) ->
    {<<>>, done};
slow_response(7) ->
    this:doesnt(exist);
slow_response(N) ->
    io:format("~p~n", [N]),
    timer:sleep(1000),
    {<<"a">>, fun() -> slow_response(N-1) end }.
```

The response you will see begins with a normal 200 OK block of headers, followed by 4 chunks, then a 500 Internal Server Error block of headers, followed by the same 4 chunks, then another 500, this time followed by the expected descriptive HTML message.

[...]

PR #333, opened 2023-02-24
stop and close connection on body-sending error

beerriot:
If an error occurs while sending the body of the response, just stop and close the connection. This prevents the problem reported in issue #60, where an error in a streaming response body causes the stream to be evaluated twice (the second time came after the error handler was engaged, and returned the same stream-body definition, because it couldn't tell whether the resource rendered its own error message).

This is a little more invasive than just catching any errors from `send_body`, in order to report the number of bytes actually sent more accurately in the access logger.

While an error message is no longer included in the response body, an error is logged by the error log handler.

### DIFF
--- a/src/webmachine_error_log_handler.erl
+++ b/src/webmachine_error_log_handler.erl
@@ -127,6 +127,12 @@ format_req(error, 503, Req, _) ->
     {Path, _} = webmachine_request:path(Req),
     Reason = "Webmachine cannot fulfill the request at this time",
     ["[error] ", Reason, ": path=", Path, $\n];
+format_req(error, 500, Req, {stream_error, Reason}) ->
+    {Path, _} = webmachine_request:path(Req),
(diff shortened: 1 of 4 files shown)


beerriot:
A note relevant to the config discussion in the previous PR: log handling is another thing that would need attention to be able to spin up multiple concurrent tests with separate webmachine supervision trees.

seancribbs:
> log handling is another thing that would need attention

We might be able to take advantage of OTP's new-ish logger for this. It would have the added benefit of being more flexible and compatible with other applications, giving the operator more control over filtering and output destinations.

seancribbs (approved):

beerriot:
> take advantage of OTP's new-ish logger

I like the sound of this … though I might go ahead and submit the few additional improvements to the existing logger that I've prepared already while I investigate. If nothing else, I think they catalog and tidy what's here, so we can better evaluate a switch to something else.

PR #334, opened 2023-02-26
DRY up webmachine_error_log_handler

beerriot:
De-duplicating log file rotation, log writing, and log formatting.


seancribbs (approved):
:shipit:

PR #335, opened 2023-02-26
test the way access and errors are logged

beerriot:
These tests test the webmachine_log event interface. It uses the `test_log_handler` introduced recently to verify that `log_access` arrives for each request, and `log_error` messages arrive for each resource error.

I didn't find any other tests covering this interface, and there's a change I'd like to make (even before the OTP logging system we were just talking about).


seancribbs (approved):
😘 👌

PR #336, opened 2023-02-27
do not call log_error from error_handler; use notes

beerriot:
This is an attempt to untangle some error logging. It probably should have been two separate commits, but I didn't want to move old stuff just to replace it immediately. So, you get two changes in one.

The first change removes the `{log_error, Code, Req, Reason}` log event. Instead of directly sending such errors to the logging system, the error is now attached to the request via the "notes" feature. At the end of request processing, notes are included in the `#wm_log_data{}` record passed with the `log_access` event. The `webmachine_error_log_handler` now looks for these notes instead.

The second change moves the responsibility for recording these errors from `webmachine_error_handler` to whatever code is raising the error. I think this better separates the responsibilities of the two similarly-named modules: `webmachine_error_handler` just renders error response messages, while `webmachine_error_log_handler` handles deciding whether the error is worth logging.

### DIFF
--- a/src/webmachine_decision_core.erl
+++ b/src/webmachine_decision_core.erl
@@ -121,6 +121,7 @@ error_response(Reason) ->
 -spec error_response(webmachine_status:status_code_with_phrase(), any()) -> ok.
 error_response(CodeAndPhrase, Reason) ->
     EndTime = erlang:monotonic_time(),
+    wrcall({add_note, error, Reason}),
     {ok, ErrorHandler} = application:get_env(webmachine, error_handler),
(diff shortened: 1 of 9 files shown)


seancribbs (approved):
This will definitely have more sane behavior.

Interestingly, the notes idea feels a lot like the annotation/span event pattern used in distributed tracing.

beerriot:
> sane behavior

Let's hope so. If nothing else, it will encourage us to make sure that `log_access` always happens.

> annotation/span event pattern

Yes! :) Curiously, the notes feature was added in 2010, but I don't think I've ever seen code use it. Let's see how it holds up.

Thanks for reviewing!

PR #337, opened 2023-03-02
remove tuple module support from webmachine_request

beerriot:
OTP 20 is our only tested release that still supports it. We don't use this internally for anything other than wasting cycles constructing and destructing a `{webmachine_request, #wm_reqstate{}}` tuple. The bare `#wm_reqstate{}` record is good enough.

We'll have to drop OTP 20 if we decide to use the new logger subsystem, which was introduced in 21. Seems like a good excuse to get rid of this clutter.


seancribbs (approved):
Not sure why I didn't approve this previously. 🤷

Issue #315, opened 2023-01-27
Is x_peername backward?


beerriot:
In writing #314, I started to wonder if `lists:last/1` is the wrong call entirely. It seems like what `webmachine_request:x_peername/2` is used for is helping `webmachine_request:peer_from_peername/2` hide the details of whether the originating client is talking directly to this server, or if a proxy is forwarding the request. I imagine that most of the time we're dealing with at most one proxy, so X-Fowarded-For is a single element.

But the [MDN docs for X-Forwarded-For](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For) say that when multiple forwards happen, it's the head of the list that is the most distant client. So `x_peername` is actually returning details about the next closest proxy. That seems like an odd thing to expose. Is `x_peername` backward?

(Sidenote: We should add support for the standard [Forwarded](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) header.)

PR #338, opened 2023-03-08
Forwarding headers fixes/improvements

beerriot:
This branch started with a fix for #315. `webmachine_request:get_peer` now returns the earliest non-private IP, which is the one closest to the client, and `webmachine_request:get_sock` returns the latest non-private IP, which is the one closest to the server.

While making the first change, I noticed that IPv6 wasn't well-supported. Localhost, link-local, and private v4 IPs translated to v6 IPs are now treated as private by `get_peer` and `get_sock`.

Finally, support is added for the standardized `Forwarded` header. `X-Forwarded-For` is checked only if no answer was found in `Forwarded`.

This is all implemented based on reading specs and other guides. I don't have a proper setup to test the behaviors of real-world proxies.


seancribbs (approved):

beerriot:
Rebased to resolve conflict with tuple-module removal.

PR #339, opened 2023-03-13
Clean dialyzer errors

martinsumner:
OTP26 dialyzer appears to be a bit smarter, and spots some type mismatches and some types which are used without first being exported.

There is a particular issue with not exporting `status_code_optional_phrase/0` from the webmachine_status_code module - as any application including webmachine.hrl (e.g. https://github.com/basho/riak_kv/blob/develop-3.0/src/riak_kv_wm_utils.erl#L48) will now get a dialyzer error as this type used in the wm_reqdata record.

Dialyzer detects that there are calls to error_response/2 in the webmachine_decision_core module that are using just status_code and not a status_code_with_phrase.  `undefined` has been passed for the phrase to clear the dialyzer error - however I'm not clear that this is the right thing to do.


beerriot (approved):
> `undefined` has been passed for the phrase to clear the dialyzer error - however I'm not clear that this is the right thing to do.

This is correct. I thought I had written them to automatically "upgrade" from bare code to `{Code, undefined}`, but it looks like my unexported dialyzer types hid that I did it wrong. Thanks for the fix!

PR #340, opened 2023-09-22
webmachine_util: fix handling of -0.0 for OTP-26.1/27.0

mikpe:
Matching of floating-point zeroes will change in OTP-27 so that -0.0 will no longer match a non-negative 0.0. OTP-26.1 warns about such constructs, which in webmachine results in:

===> Compiling src/webmachine_util.erl failed
src/webmachine_util.erl:267:18: matching on the float 0.0 will no longer also match -0.0 in OTP 27. If you specifically intend to match 0.0 alone, write +0.0 instead.
src/webmachine_util.erl:270:10: matching on the float 0.0 will no longer also match -0.0 in OTP 27. If you specifically intend to match 0.0 alone, write +0.0 instead.
src/webmachine_util.erl:275:10: matching on the float 0.0 will no longer also match -0.0 in OTP 27. If you specifically intend to match 0.0 alone, write +0.0 instead.
@src/webmachine_util.erl:299:9: matching on the float 0.0 will no longer also match -0.0 in OTP 27. If you specifically intend to match 0.0 alone, write +0.0 instead.

Fixed by switching from matches to numerical comparisons.

### DIFF
--- a/src/webmachine_util.erl
+++ b/src/webmachine_util.erl
@@ -264,15 +264,15 @@ do_choose(Choices, Header, Default) ->
     DefaultOkay = case DefaultPrio of
         [] ->
             case StarPrio of
-                [0.0] -> no;
+                [SP] when SP == 0.0 -> no;
                 _ -> yes
             end;
-        [0.0] -> no;
+        [DP] when DP == 0.0 -> no;
         _ -> yes
     end,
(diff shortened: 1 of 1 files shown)


beerriot (approved):
Looks like the right fix to me. Thanks!

PR #342, opened 2023-10-21
update mochiweb dep to get upstream 0.0 matching fix

beerriot:
mochi/mochiweb#260 is the same sort of fix as webmachine/webmachine#340 - updating floating point 0.0 matching for OTP 27 compatibility


seancribbs (approved):
Nice!

PR #344, opened 2024-02-18
Fix update float format for 27+

asyncmind0:


beerriot:
Thanks for the patch, @jagguli. But I think we're all set here. We don't care whether it's +0.0 or -0.0, and `== 0.0` will get us `true` in either case. (https://www.erlang.org/doc/general_info/upcoming_incompatibilities#0.0-and--0.0-will-no-longer-be-exactly-equal)

Please add more information about what issues you're seeing, and how this fixes them if I've missed something, though.

asyncmind0:
I was getting this warning , I guess I can workaround by disabling warning_as_error in rebar.config , I'll update with more info soon
```
To help to find code that might need to be revised, in OTP 27 there will be a new compiler warning when matching against 0.0 or comparing to that value using the =:= operator. The warning can be suppressed by matching against +0.0 instead of 0.0.
```

warning message .. 

```
 matching on the float 0.0 will no longer also match -0.0 in OTP 27. If you specifically intend to match 0.0 alone, write +0.0 instead.
```

I've tried to re read that section to make sure I'm getting it right ,but my brains stack overflowing...

Im on OTP 26.2.2 and beam 14.2.2, on Arch Linux, so probably they implemented the warning they mentioned

```
We plan to introduce the same warning in OTP 26.1, but by default it will be disabled.

```
It should be disabled by default tho ... hmm

beerriot:
We did see this warning on earlier earlier OTP 26 builds, but we fixed them in #340 and #342. Please check that you're at commit 7bdd27d (the current HEAD on main). The specific use of the term "matching" makes me suspect you may be using a commit before that point (`when SP == 0.0` is not a "match" in Erlang terms, but the older code `[0.0] ->` is).

PR #346, opened 2024-12-04
update mochiweb dep to get upstream 'maybe' fix

beerriot:
OTP 27 enables the new `maybe` keyword by default. Without setting compiler flags, a Webmachine project will fail to build with the compiler claiming that `{maybe, Start}` in mochiweb is a syntax error.


seancribbs (approved):
🚀

Issue #347, opened 2025-03-20
Maybe time for a new version tag?


tburghart:
The last tagged version is from 2018, and a lot has been updated.
Perhaps a new version tag is in order?

beerriot:
This does sound like a good idea. If I have ever been the person to bump&tag this repo, though, it has been a LONG time. So you'll have to bear with me if I goof.

@seancribbs @joedevivo @justinsheehy … anyone have an unused quip stashed somewhere?

justinsheehy:
Have we ever used "Stuff just happened." as a quip? It's from the same source material as the current quip.

(I have not exhaustively looked through history to verify whether it has ever been used)

beerriot:
> Stuff just happened.

Not in the history that I can find. Thanks! Making it so…

beerriot:
🎉 https://github.com/webmachine/webmachine/releases/tag/1.12.0

PR #348, opened 2025-03-21
Version bump (for tagging 1.12.0)

beerriot:
As proposed in #347. I'll tag 1.12.0 after merge?

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -14,7 +14,7 @@
 
 {profiles,
  [{test,
-   [{deps, [meck,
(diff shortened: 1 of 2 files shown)


beerriot:
Looks like CI is broken. First time I've released on a Friday in a while. 🤣

beerriot:
Meck released 1.0.0 since our last change, which added support for OTP 27, but dropped support for OTP before 25. Let's err on releasing what we've had, and do another bump to update ourselves to newer OTP when we're ready?

seancribbs (approved):

PR #349, opened 2026-07-14
Fix maybe_flush_req_body/1 hang when body was eagerly read

samwar:
## Summary

Fixes a POST/PUT-request hang introduced in 1.12 where the new
`wrcall(maybe_flush_req_body)` step in
`webmachine_decision_core:finish_response/2` unconditionally attempts to
drain the socket with an infinite recv timeout, even when the request
body was already fully consumed via `wrq:req_body/1`.

## Root cause

`maybe_flush_req_body/1` was introduced in 1.12 to drain any request-body
bytes still on the socket before `send_response` runs. Its `undefined`
branch of `get_stream_progress/0` unconditionally called `recv_stream_body`
with `?IDLE_TIMEOUT = infinity`, which is unsafe in three ways:

1. Resources that read the request body eagerly (`wrq:req_body/1`) set
   `mochiweb_request_recv` but never touch `webmachine_stream_progress`.
   `maybe_flush_req_body` sees `undefined`, attempts to drain, and blocks
   forever on an already-empty socket. The request completes on the
   server side but the response is never written and the client hangs
   until its own timeout. This regresses correctness for every resource
   that uses the eager-read path.

2. Applications that dispatch body reads to a separate worker process
   (a common pattern for offloading long-running work) get
   `mochiweb_request_recv` set on the worker's pdict rather than on
   the mochiweb loop process that runs `maybe_flush_req_body/1`. The
   loop process still sees `undefined` for both `webmachine_stream_progress`
   and `mochiweb_request_recv` and falls into the drain path even though
   the body was fully consumed.

3. Even for resources that legitimately reach the drain path, using
   `?IDLE_TIMEOUT = infinity` for a fallback flush is a footgun: a
   misbehaving client that half-closes the write side and sends a
   Content-Length header but no body wedges the request pipeline
   indefinitely.

## Fix

Three defenses, in priority order:

* **Fast path (pdict guard).** Guard the `undefined` branch on
  `get(mochiweb_request_recv) =:= true` and return `true` immediately in
  that case. Eager reads mark the request as recv'd; there's nothing
  left on the socket to drain.

* **Cross-process fallback (ReqState guard).** When the pdict flag is
  not set, consult `Req#wm_reqstate.bodyfetch`. `bodyfetch` is set to
  `standard` or `stream` inside `call({req_body, ...}, ReqState)` in
  the same call path that would set the pdict flag, and unlike the
  pdict it travels with the `ReqState` value across process boundaries.
  Applications that thread the updated `ReqState` back to the mochiweb
  loop process (via the returned `Request#wm_reqdata.wm_state` field)
  will hit this fallback and short-circuit the drain.

* **Safety net (`?FLUSH_TIMEOUT`).** When neither signal is set, still
  attempt to drain, but with `?FLUSH_TIMEOUT = 5000` instead of
  `?IDLE_TIMEOUT = infinity`. Threaded through a new
  `recv_stream_body/3` variant used only by `maybe_flush_undefined_drain/1`.
  Legitimate streaming callers (`stream_req_body`, `do_recv_body`) still
  get `?IDLE_TIMEOUT = infinity` via the 2-arg wrapper.

## Test evidence

Reproduced against OTP 26.2.5 in a downstream application. A large
POST/PUT eunit suite hung indefinitely on 1.12.0 for every resource
that used `wrq:req_body/1`. After the patch, target-class cases pass.
For a further subset that used a cross-process worker pattern, the
`?FLUSH_TIMEOUT` safety net converted the infinite hang into a 5s
per-request delay; the additional `bodyfetch` fallback resolves those
by consulting `ReqState` in addition to the pdict.

Compact dbg trace showing the original hang shape:

```
18:31:58.164 maybe_flush_req_body ENTRY progress=undefined
18:31:58.164 maybe_flush_req_body branch=undefined expects_continue=false
18:31:58.164 maybe_flush_req_body undefined->drain max_flush=67108864
18:31:58.164 maybe_flush_req_body BEFORE recv_stream_body read_size=65535
18:31:58.164 mochiweb_socket:recv ENTRY sock=tcp length=21 timeout=infinity
18:32:02.302 [context murdered by client timeout, no AFTER log]
```

## Compatibility

Additive — no existing callers of `recv_stream_body/2` change semantics.
A new `recv_stream_body/3` variant is added for the finite-timeout
fallback path. Legitimate streaming reads (`stream_req_body`,
`do_recv_body`) still use `?IDLE_TIMEOUT = infinity`. The new `bodyfetch`
guard is a pure lookup on an existing `#wm_reqstate{}` field; no new
fields introduced, no message-shape changes.

### DIFF
--- a/src/webmachine_request.erl
+++ b/src/webmachine_request.erl
@@ -87,6 +87,12 @@
 
 -define(IDLE_TIMEOUT, infinity).
 
+%% Finite timeout used when we're only draining leftover request bytes
+%% (see maybe_flush_req_body/1). Kept intentionally short so that a
+%% misbehaving or already-drained peer can never wedge the request
+%% pipeline. Legitimate streaming reads still use ?IDLE_TIMEOUT.
+-define(FLUSH_TIMEOUT, 5000).
+
 -type t() :: #wm_reqstate{}.
 -export_type([t/0]).
 
@@ -588,7 +594,10 @@ expects_continue(PassedState) ->
 sent_continue() ->
     get(webmachine_sent_continue) =:= true.
 
-recv_stream_body(PassedState=#wm_reqstate{reqdata=RD}, MaxHunkSize) ->
+recv_stream_body(PassedState, MaxHunkSize) ->
+    recv_stream_body(PassedState, MaxHunkSize, ?IDLE_TIMEOUT).
+
+recv_stream_body(PassedState=#wm_reqstate{reqdata=RD}, MaxHunkSize, Timeout) ->
     case expects_continue(PassedState) and (not sent_continue()) of
         true ->
             put(webmachine_sent_continue, true),
@@ -608,28 +617,29 @@ recv_stream_body(PassedState=#wm_reqstate{reqdata=RD}, MaxHunkSize) ->
             {<<>>, done};
         chunked ->
             start_recv_chunked_body(PassedState#wm_reqstate.socket,
-                                    MaxHunkSize);
+                                    MaxHunkSize, Timeout);
         Length ->
             recv_unchunked_body(PassedState#wm_reqstate.socket,
-                                MaxHunkSize, Length)
+                                MaxHunkSize, Length, Timeout)
     end.
 
-recv_unchunked_body(Socket, MaxHunk, DataLeft) ->
+recv_unchunked_body(Socket, MaxHunk, DataLeft, Timeout) ->
     case MaxHunk >= DataLeft of
         true ->
-            case mochiweb_socket:recv(Socket,DataLeft,?IDLE_TIMEOUT) of
-                {ok,Data1} ->
+            case mochiweb_socket:recv(Socket, DataLeft, Timeout) of
+                {ok, Data1} ->
                     record_stream_progress(done),
                     {Data1, done};
                 {error, Error} ->
                     throw({webmachine_recv_error, Error})
             end;
         false ->
-            case mochiweb_socket:recv(Socket,MaxHunk,?IDLE_TIMEOUT) of
-                {ok,Data2} ->
+            case mochiweb_socket:recv(Socket, MaxHunk, Timeout) of
+                {ok, Data2} ->
                     Next = fun() ->
                                    recv_unchunked_body(Socket, MaxHunk,
-                                                       DataLeft-MaxHunk)
+                                                       DataLeft - MaxHunk,
+                                                       Timeout)
                            end,
                     record_stream_progress(Next),
                     {Data2, Next};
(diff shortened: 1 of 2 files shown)


beerriot:
Thanks for the patch, @samwar! Would it be possible to distill one of the downstream-app test failures you found into a test that we can add here? I'm especially interested to have one that covers Root Cause 2 - handing the stream reading off to another process. That sounds like an interesting case that I think would be good to document here.

I'll admit that I'm also curious why the `post_then_get` test in `multi_request_connection_test`, or a few of the `decision_core_test` cases don't cover Root Cause 1. They each seem to call `wrq_req:body` from their POST handlers, with the test checking what response the client got. How is the call in this problem case different?

Agreed that an infinite timeout on reading what's left on the socket is a bad idea, in general, though. +1 to this patch resolving that.

samwar:
Good questions @beerriot! I dug into `post_then_get` before writing the test to make sure I understood why the existing suite lets the hang slip through.

**On why `post_then_get` (and the `decision_core_test` POST cases) don't hit the hang:**

They pipeline two requests on the same socket. When `maybe_flush_req_body/1` runs after Req1 (POST), Req2's bytes are already sitting on the socket. But it doesn't matter, because the eager `wrq:req_body/1` in `process_post/2` runs `recv_stream_body/2` under `do_recv_body/1`, which calls `record_stream_progress(done)`. `maybe_flush_req_body/1` then finds `webmachine_stream_progress = done` in the pdict and short-circuits before ever touching the socket. The follow-up request bytes are incidental — the machinery just never asks.

The hang lands when the eager read is dispatched to a **worker process**, which is a pattern real downstream callers use. `wrq:req_body/1` records `webmachine_stream_progress = done` and `req_body` in the worker's pdict, not the loop's. The updated ReqState (with `bodyfetch = standard`) is returned to the caller via `put(tmp_reqstate, ...)` in the worker's pdict. If the resource propagates that back to the loop process's pdict, which `webmachine_resource:do/2` picks up on callback return. The loop has a ReqState with `bodyfetch = standard`, but its own pdict has no `webmachine_stream_progress`. On pre-fix code, `maybe_flush_req_body/1` sees `undefined` and falls into `recv_stream_body(Req, ReadSize)` with `?IDLE_TIMEOUT = infinity`, blocking on a socket that has nothing left. A single-request client (curl, HTTP libraries, load balancers) is now stuck waiting for a response the server will never send. That's what took our downstream service down.

I've pushed [`test/flush_req_body_test.erl`](https://github.com/raxdcx/webmachine/blob/flush-req-body-guard/test/flush_req_body_test.erl) with one test, `cross_process_eager_read_test`, that reproduces this. `process_post/2` spawns a worker that calls `wrq:req_body/1`, receives the worker's `tmp_reqstate` back, and stores it in the loop's pdict (mirroring the downstream pattern). The client is single-request-per-socket with a 3-second `gen_tcp:recv` timeout.

**Verified differential** (full `rebar3 eunit` on OTP 26, 179 tests total):

- Post-fix branch: 179 pass, response returns in ~1s. Suite finishes in ~10.4s.
- Pre-fix branch (`upstream/main`) with just the test file added: 178 pass, `cross_process_eager_read_test` fails with `{error, timeout}`. Suite finishes in ~13.7s — the ~3.3s extra is the client's recv timeout burning while the server is stuck in `mochiweb_socket:recv/3` at `?IDLE_TIMEOUT = infinity`.

One correction from my earlier draft: I initially wrote up a second test aimed at the "eager read in the loop process" case (pdict guard). When I actually ran the diff, that shape doesn't hang on pre-fix code. The same-process eager reads set `webmachine_stream_progress = done` and hit the existing `done` branch cleanly. The pdict guard in the patch is defense-in-depth for the (theoretical) state where `mochiweb_request_recv = true` but `webmachine_stream_progress = undefined`; I don't have a way to reach it via normal callbacks, so no there's regression test for it. The `Req#wm_reqstate.bodyfetch` fallback is the one that actually closes the hole downstream, and it's the one this test exercises.

Happy to rename, split, or reshape if you'd like the test to look more like the existing multi_request_connection cases.

beerriot (approved):
I forgot that I wanted to mention #332 here, to provide a link to the change in v1.12 that added the flush getting fixed in this issue.

But thanks for the explanation, and for the test, @samwar. Approved, pending the rest of the test runs finishing!

samwar:
@beerriot Some of those tests have been pending completion for a while now. Before I submitted my fix, the previous commit had tests that had been waiting to run for several days. How long is the process wait supposed to take? The tests that are pending also running against versions of OTP that have been deprecated. Is it worth cutting those versions off for the newer versions of webmachine?

beerriot:
I've just taken a brief look, and I bet the reason those test runs are stuck is that they're trying to run on ubuntu 20.04, which appears to no longer be an os version supported by the standard github action runners. I have no opposition to dropping them. It would be good to add cases for the newer OTP versions as well. Let's see if I can fix things quickly here…

beerriot:
Okay, test updates have made it to main. If you don't mind rebasing this branch, I expect it will pass afterward, and we can get it merged. Thanks!

samwar:
@beerriot I rebased and force pushed to get the test changes. The required checks will need to be updated in the branch options. They still look for old jobs and expect them to pass before anything can be merged.

beerriot:
> The required checks will need to be updated in the branch options.

@seancribbs I don't have the admin permissions to see/set these.

seancribbs:
@beerriot Fixed and made you admin 😄

PR #351, opened 2026-07-21
drop testing of OTP versions 20-23

beerriot:
Their github actions requested ubuntu 20.04 images, which standard runners no longer support. OTP 23 is six years old, so I think it's safe to move on.

I think I remember last time we updated this file, we had to push directly to main, instead of going this route, because github actions will pick up the old build-test.yaml to test the new branch. But let's see what happens.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -6,7 +6,7 @@ jobs:
   build:
     strategy:
       matrix:
-        otp_version: [25, 24]
+        otp_version: [28, 27, 26, 25, 24]
         rebar3_version: ["3.20"]
         os: ["ubuntu-latest"]
       fail-fast: false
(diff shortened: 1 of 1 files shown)


beerriot:
Dropping old OTPs work, except for an expected OTP 23 run hanging things up. I'm amending this PR to also fix two other things:

1. Node 20 being deprecated - it looks like just updating the checkout action version should resolve the warning about it?
2. Adding OTP versions we've wanted to test for a while - it looks like OTP 26 is easy, but 27-29 likely require a rebar update.

seancribbs (approved):

beerriot:
Okay, we're testing OTP 24-28 now. OTP 29 can be added in the same manner (at rebar version 3.27), but it currently reports some new warnings that need to be addressed before the compile step will succeed.

@seancribbs any chance you're around for a 👍🏻?

beerriot:
@seancribbs I think we may need you to merge this one. The main branch is protected, and the checks are still picking up old definitions that are stuck. I'm unable to get to the repository settings to override the branch protection.

(And thanks for the merge approval!)

beerriot:
tyvm!

